PR #3, opened 2025-07-28
Create php-developer

Donubu:
Added php-developer claude code agent.


davepoon:
Thanks for the contribution 😄

PR #133, opened 2026-04-23
Add claude Super Init plugin

apyx-dev:


davepoon (changes requested):
**Security/content review**

The skill/agent content looks like normal workflow scaffolding (explicit role preamble, phased steps, sequential execution). Nothing stood out as a hidden “ignore policy” jailbreak layered for users who open the Markdown.

**Blocking issue**

`plugins/super-init/.claude-plugin/plugin.json` is **invalid strict JSON**: there is a **trailing comma** after `"automation"` in the `keywords` array (line 10). That will fail many validators and parsers.

Please remove that trailing comma and push; happy to re-check after.

davepoon:
Closing this one per maintainer policy: ** isn’t strict JSON** (trailing comma after the last `keywords` entry), so tooling and validators can’t rely on parsing it reliably. Once that comma is fixed, feel free to re-open or send a fresh PR.

Separately—and easy to soften in wording—the `super-init` command text tells the assistant not to confirm with you before executing; we prefer commands that keep the human in control for destructive or environment-changing flows.

Thanks again for the detailed init skill and tooling work—we’d gladly take another look after the manifest is valid.

davepoon:
Quick correction to my closing note above (formatting glitch): invalid strict JSON is in `plugins/super-init/.claude-plugin/plugin.json` — trailing comma after the final `keywords` array element (`"automation",]`). Removing that comma should make the manifest parse everywhere.

PR #144, opened 2026-05-11
feat(hooks): add sql-bulk-delete-warn for destructive SQL detection

yurukusa:
A new hook at \`plugins/all-hooks/hooks/sql-bulk-delete-warn.md\` that warns when a destructive SQL statement (\`DELETE\` / \`UPDATE\` / \`TRUNCATE\`) is about to run via \`psql\`, \`mysql\`, \`sqlite3\`, \`sqlcmd\`, \`cockroach sql\`, or \`duckdb\` without a row-count safeguard.
This fills a gap in the current \`all-hooks\` bundle, which has \`file-protection.md\` but no SQL-specific safety net.
The hook is motivated by [claude-code issue #56738](https://github.com/anthropics/claude-code/issues/56738) (2026-05-06): a regex with capture groups in \`regexp_match\` returned \`NULL\` for nearly every row, and the subsequent \`DELETE FROM table WHERE _clean_title IS NULL\` wiped **24,472 of 24,475 rows**. \`autovacuum\` cleaned the dead tuples within ~5 minutes, blocking \`pg_dirtyread\` recovery. Three days of scraping work permanently lost. The exact pattern is detectable at \`PreToolUse\` time, before the SQL ever reaches the database.
1. \`DELETE\` / \`UPDATE\` with no \`WHERE\` clause (full-table touch)
2. \`DELETE WHERE col IS NULL\` — the #56738 template: a \`NULL\` arising from a computed column populated by a prior \`UPDATE\` that may have failed silently
3. \`TRUNCATE TABLE\` (non-recoverable in some engines)
4. \`psql -c\` with destructive SQL but no \`BEGIN\` / \`START TRANSACTION\` wrapping (autocommit persists before the user can verify)
Advisory by default. Setting \`CC_SQL_BULK_DELETE_BLOCK=1\` converts the warning into a hard block (\`exit 2\`).
The detection logic comes from [cc-safe-setup](https://github.com/yurukusa/cc-safe-setup) (MIT). Contributed here as a single-file hook per this repo's \`all-hooks\` convention. Source attribution is included at the bottom of the \`.md\` file. No reciprocal dependency is introduced — buildwithclaude users can install the hook standalone.
- [ ] hook detects \`psql -c "DELETE FROM users"\`
- [ ] hook detects \`mysql -e "DELETE FROM logs WHERE id IS NULL"\`
- [ ] hook ignores \`psql -c "SELECT * FROM users"\`
- [ ] hook ignores \`psql -c "DELETE FROM users WHERE id = 1"\` (has WHERE with specific value)
- [ ] \`CC_SQL_BULK_DELETE_BLOCK=1\` causes exit 2; default is advisory (exit 0)

### DIFF
--- a/plugins/all-hooks/hooks/sql-bulk-delete-warn.md
+++ b/plugins/all-hooks/hooks/sql-bulk-delete-warn.md
@@ -1,0 +1,101 @@
+---
+name: sql-bulk-delete-warn
+description: Warn when a destructive SQL command (DELETE/UPDATE/TRUNCATE) runs via psql, mysql, sqlite3, or sqlcmd without a row-count safeguard
+category: security
+event: PreToolUse
+matcher: Bash
+language: bash
+version: 1.0.0
+---
+
+# sql-bulk-delete-warn
+
+Warn when a destructive SQL command (`DELETE` / `UPDATE` / `TRUNCATE`) is about to run via `psql`, `mysql`, `sqlite3`, `sqlcmd`, `cockroach sql`, or `duckdb` without a row-count safeguard.
+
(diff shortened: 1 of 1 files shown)


davepoon:
✅ `main` now includes [#149](https://github.com/davepoon/buildwithclaude/pull/149) (marketing-optimizer CI fix + `windows-python-stub-detector` with valid hook schema). Rebase or **Update branch** from `main` if checks were red; then this PR should go green.

PR #145, opened 2026-05-11
feat(hooks): add windows-python-stub-detector for silent Python hook no-op on Windows Git Bash

yurukusa:
## What this adds

A new hook at `plugins/all-hooks/hooks/windows-python-stub-detector.md` that surfaces the Microsoft Store `python3` stub on Windows Git Bash at `SessionStart`, before any Python-based hook silently no-ops.

## Why

This addresses [claude-code issue #57946](https://github.com/anthropics/claude-code/issues/57946) — a failure mode that is silent by construction: every Python-based hook (PreToolUse, PostToolUse, etc.) appears to run on Windows Git Bash, but actually exits 49 with no stdout/stderr before any inspection logic runs. The operator sees a valid hook configuration, a real `.py` file on disk, and `which python3` succeeding — and concludes the hook is active. The hook is not active. Every tool call passes uninspected.

The hook detects four variants of the failure at session start and emits a `hookSpecificOutput.additionalContext` warning the model can relay to the user. No blocking — `SessionStart` blocks would prevent the user from working at all.

## How it works

Probes the binary at `$CC_PYTHON_CMD` (default `python3`) with a trivial Python statement. Trips the warning if any of:

1. **Exit 49** — the documented Store-redirect signal in Git Bash subprocess.
2. **Stderr contains "Microsoft Store" / "install Python"** — the Store-redirect marker some shell wrappers expose.
3. **Exit 127 or "command not found"** — `python3` missing from PATH (separate-but-adjacent failure).
4. **Exit 0 but no "ok" in output** — the silent stub variant that exits cleanly without running the script. This is the worst case: every other safety net assumes exit 0 means success.

## Why this isn't a duplicate

- `file-protection.md` covers protected files; nothing in the current `all-hooks` set probes the runtime.
- No `SessionStart` hooks currently ship in `all-hooks` — this is the first one. The `SessionStart` event is exactly the right place for environment diagnostics: it runs before any tool call, so a stale runtime is caught before it can fail silently downstream.

## Attribution

The detection logic comes from [cc-safe-setup](https://github.com/yurukusa/cc-safe-setup) (MIT). Contributed here as a single-file hook per this repo's `all-hooks` convention. Source attribution at the bottom of the `.md` file. No reciprocal dependency — buildwithclaude users on Windows Git Bash can install standalone.

## Test plan

- [ ] Hook returns `hookSpecificOutput` JSON when `CC_PYTHON_CMD=/path/to/fake-stub` (exit 49 simulation)
- [ ] Hook returns nothing (exit 0) when `python3` works normally
- [ ] Hook detects exit 127 / command not found
- [ ] Hook detects silent stub variant (exit 0, empty stdout)
- [ ] No blocking — exit code always 0
- [ ] CI passes

### DIFF
--- a/plugins/all-hooks/hooks/windows-python-stub-detector.md
+++ b/plugins/all-hooks/hooks/windows-python-stub-detector.md
@@ -1,0 +1,92 @@
+---
+name: windows-python-stub-detector
+description: SessionStart probe that surfaces the Microsoft Store python3 stub on Windows Git Bash before Python-based hooks silently no-op
+category: diagnostic
+event: SessionStart
+matcher: ""
+language: bash
+version: 1.0.0
+---
+
+# windows-python-stub-detector
+
+SessionStart probe that surfaces the Microsoft Store `python3` stub on Windows Git Bash before Python-based hooks silently no-op.
+
+Motivated by [claude-code issue #57946](https://github.com/anthropics/claude-code/issues/57946). On Windows Git Bash, `python3` may resolve to the Microsoft Store stub rather than a real interpreter. The stub is on `PATH`, `which python3` succeeds, but invoking it as a subprocess exits 49 with no stdout and no stderr. Python-based `PreToolUse` and `PostToolUse` hooks then exit silently before they can inspect the tool call — the hook configuration is valid, the script file is present, and the agent proceeds as if the hook approved. This is the silent-no-op failure mode that gives every operator the false belief their hooks are running.
+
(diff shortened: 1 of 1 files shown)


davepoon:
Thank you — closing as superseded by [#149](https://github.com/davepoon/buildwithclaude/pull/149), which includes the same `windows-python-stub-detector` hook with **schema-compliant** frontmatter (`category: development`, non-empty `matcher: "*"` per `scripts/hook-schema.json`). Your attribution and body text are preserved there. If you want further edits, open a small follow-up PR on `main`.

PR #158, opened 2026-05-20
fix(indexer): paginate topic search to mirror code-search pagination

### LINKED ISSUE #157: Topic-search discovery is single-page-only — small/new marketplaces are silently dropped
## Summary

The marketplace indexer's topic-search loop in `web-ui/lib/indexer/marketplace-indexer.ts` calls `github.searchRepositories()` exactly once per topic and never paginates. Because `searchRepositories` returns at most 100 results per call (sorted by `stars,desc`), any repo ranked past position #100 for that topic is silently dropped — even though the helper accepts a `page` parameter and the parallel code-search block right above it does paginate correctly.

## Reproduction

`topic:claude-code-plugins` currently has 266 matching repositories — well over the 100/page limit. The repos at ranks 101–266 are never seen by the indexer.

```bash
# Total repos carrying the topic
gh api 'search/repositories?q=topic:claude-code-plugins&per_page=1' --jq '.total_count'
# → 266

# What the indexer sees today (page 1 only, the buggy path)
gh api 'search/repositories?q=topic:claude-code-plugins&sort=stars&order=desc&per_page=100&page=1' \
  --jq '.items | length'
# → 100  (166 silently dropped)
```

Concrete example — my own org's three repos all carry `topic:claude-code-plugins`, and the marketplace one has a valid `.claude-plugin/marketplace.json`:

| Repo | Star rank in topic | In repoSet today? |
|---|---|---|
| `glitchwerks/claude-prospector` | #172 | ❌ |
| `glitchwerks/plugins` (marketplace) | #195 | ❌ |
| `glitchwerks/claude-wayfinder` | #201 | ❌ |

Local test of the proposed fix — paginate to `Math.min(Math.ceil(total/100), 10)` pages, mirroring the existing code-search loop:

```bash
for p in 1 2 3; do
  gh api "search/repositories?q=topic:claude-code-plugins&sort=stars&order=desc&per_page=100&page=$p" \
    --jq '.items[].full_name'
done | sort -u | grep glitchwerks
# →
#   glitchwerks/claude-prospector
#   glitchwerks/claude-wayfinder
#   glitchwerks/plugins
```

All three would enter the repoSet. `glitchwerks/plugins` would then pass `isMarketplace()` (`claude-code-plugins` is in `marketplaceTopics`) and `parseMarketplaceJson()` (file exists at the expected path), and upsert into the `marketplaces` table on the next cron run.

## Root cause

`web-ui/lib/indexer/marketplace-indexer.ts` around L113–125:

```ts
// Topic-based repository search
const topicQueries = [
  'claude-code-skill',
  'claude-skill',
  'agent-skill',
  'claude-agent-skill',
  'claude-code-plugins',
]

for (const topic of topicQueries) {
  try {
    const { repos } = await github.searchRepositories(`topic:${topic}`)  // page 1 only
    console.log(`Found ${repos.length} repos for topic: ${topic}`)
    for (const repo of repos) {
      repoSet.add(repo.full_name)
    }
  } catch (error) {
    console.error(`Topic search failed for "${topic}":`, error)
  }
}
```

Compare to the **code-search** block immediately above it (~L88–105), which paginates correctly:

```ts
const searchResult = await github.searchCode(searchQuery)
const maxPages = Math.min(Math.ceil(searchResult.total_count / 100), 10)
for (let page = 1; page <= maxPages; page++) {
  const pageResult = page === 1 ? searchResult : await github.searchCode(searchQuery, page)
  for (const item of pageResult.items) {
    repoSet.add(item.repository.full_name)
  }
}
```

Supporting signals this is a bug rather than an intentional cap:

- `searchRepositories(query: string, page: number = 1)` exposes a `page` parameter that the caller never uses.
- The helper also returns `totalCount`, which is only useful for *deciding how to paginate*. The caller discards it.
- No commit message in the file's history mentions popularity, threshold, cap, throttle, or filter. The single-page form has existed since the file was first committed (8ed0298 → 224286d, 2026-01-08) and has never been touched.
- No issue or PR references a popularity gate.
- README and CONTRIBUTING.md document no minimum-star or minimum-age policy.
- When this codebase intentionally filters something — e.g. the March 2026 admin-review `submissionStatus` gate (#commit 8768d36) — the change is named, schema'd, and audit-logged. The topic-search cap has none of those fingerprints.

---

🤖 Generated by Claude Code on behalf of @cbeaulieu-gt

cbeaulieu-gt:
Fixes #157.

## Summary

The topic-search loop in `indexMarketplaces` (`web-ui/lib/indexer/marketplace-indexer.ts`) calls `github.searchRepositories(topic)` exactly once and never paginates — silently dropping any repo ranked past position #100 in `searchRepositories`'s `stars,desc` ordering. The helper exposes a `page` parameter and returns `totalCount`; both were unused at the call site.

This PR mirrors the pagination pattern from the code-search block immediately above it in the same function:

```ts
const maxPages = Math.min(Math.ceil(searchResult.total_count / 100), 10)
for (let page = 1; page <= maxPages; page++) {
  const pageResult = page === 1 ? searchResult : await github.searchCode(searchQuery, page)
  ...
}
```

## The diff

```diff
 for (const topic of topicQueries) {
   try {
-    const { repos } = await github.searchRepositories(`topic:${topic}`)
-    console.log(`Found ${repos.length} repos for topic: ${topic}`)
-    for (const repo of repos) {
-      repoSet.add(repo.full_name)
+    const first = await github.searchRepositories(`topic:${topic}`)
+    console.log(`Found ${first.totalCount} repos for topic: ${topic}`)
+    const maxPages = Math.min(Math.ceil(first.totalCount / 100), 10)
+    for (let page = 1; page <= maxPages; page++) {
+      const pageResult = page === 1 ? first : await github.searchRepositories(`topic:${topic}`, page)
+      for (const repo of pageResult.repos) {
+        repoSet.add(repo.full_name)
+      }
     }
   } catch (error) {
     console.error(`Topic search failed for "${topic}":`, error)
```

## Why `Math.min(..., 10)`

Same ceiling as the existing code-search loop, and matches GitHub's hard 1000-result cap on Search queries (page 11 returns HTTP 422). Not policy — just the API limit.

## API-call cost

Today this means +2 calls per cron run (`claude-code-plugins` has 266 results → 3 pages). The other four topics (`claude-code-skill`, `claude-skill`, `agent-skill`, `claude-agent-skill`) likely fit in one page each, so they're unchanged. The throttle/backoff/circuit-breaker in `GitHubClient` already covers the additional load.

## Test plan

- [ ] Trigger the indexer manually (cron route or admin endpoint) and confirm `Found N repos for topic: ...` now logs the full totalCount for `claude-code-plugins`
- [ ] After the run, query `/api/marketplaces/list?search=glitchwerks` (or any small marketplace previously past rank #100 in `claude-code-plugins`) and confirm it appears

I was unable to test against your Postgres/Neon DB locally, so the runtime verification is gated on the cron trigger in your environment. Happy to iterate on the shape if you'd prefer a different pagination idiom or want telemetry around the new pages.

---

🤖 Generated by Claude Code on behalf of @cbeaulieu-gt

### DIFF
--- a/web-ui/lib/indexer/marketplace-indexer.ts
+++ b/web-ui/lib/indexer/marketplace-indexer.ts
@@ -120,10 +120,14 @@ export async function indexMarketplaces(): Promise<IndexResult> {
 
   for (const topic of topicQueries) {
     try {
-      const { repos } = await github.searchRepositories(`topic:${topic}`)
-      console.log(`Found ${repos.length} repos for topic: ${topic}`)
-      for (const repo of repos) {
-        repoSet.add(repo.full_name)
+      const first = await github.searchRepositories(`topic:${topic}`)
+      console.log(`Found ${first.totalCount} repos for topic: ${topic}`)
+      const maxPages = Math.min(Math.ceil(first.totalCount / 100), 10)
+      for (let page = 1; page <= maxPages; page++) {
+        const pageResult = page === 1 ? first : await github.searchRepositories(`topic:${topic}`, page)
+        for (const repo of pageResult.repos) {
+          repoSet.add(repo.full_name)
+        }
       }
     } catch (error) {
       console.error(`Topic search failed for "${topic}":`, error)


cbeaulieu-gt:
Ran `tsc --noEmit` against `web-ui/tsconfig.json` locally with the diff applied — passes clean, exit 0, zero errors. Confirms the field-name choices (`first.totalCount`, `first.repos`, `pageResult.repos`) line up with `GitHubClient.searchRepositories`'s return type.

That's the only static gate available in this repo (no test runner / no `*.test.*` files / no `vitest`/`jest` deps); runtime validation against the Postgres/Neon DB is still up to you per the test plan above.

🤖 Generated by Claude Code on behalf of @cbeaulieu-gt

cbeaulieu-gt:
Functional check — wired the project's real `GitHubClient` to the new pagination pattern and ran it against the live GitHub API. Confirms the buggy behavior is reproducible and the fix surfaces a previously-missed marketplace.

```
[pre-fix]   page-1-only: 100 repos
[pre-fix]   contains glitchwerks/plugins? false   ← reproduces the bug

[fixed]   totalCount=266
[fixed]   maxPages=3
[fixed]   page 1: +100 repos (total 100)
[fixed]   page 2: +100 repos (total 200)
[fixed]   page 3: +66 repos (total 266)

[fixed]   contains glitchwerks/plugins? true    ← PR fixes it
[fixed]   final repoSet size: 266
```

The harness imports `getGitHubClient` from `web-ui/lib/github/client.ts` and exercises the *exact* loop body from the PR diff — same `Math.min(Math.ceil(first.totalCount / 100), 10)` formula, same `page === 1 ? first : fetch(page)` pattern, same field names. So the only thing not covered by this signal is the DB upsert path, which is gated on your Postgres/Neon env anyway and is unchanged by this PR.

Sanity caveat: `glitchwerks/plugins` is at star-rank #195 today and only #195 because the indexer dropping it kept it at zero stars (and #172/#201 are sibling repos in the same org with the same topic). If the rank shifts above #100 between now and merge, the harness's *negative* case for the buggy path would still hold for the other 166 dropped repos; only the specific repo name in the assertion would need updating.

Net signal stack on this PR: ✅ typecheck (`tsc --noEmit`), ✅ runtime harness against the live API. Project ESLint config has a pre-existing circular-reference crash during config validation (`eslint-config-next@16.1.6` interaction with eslintrc compat layer) — fires before any file is examined, so not informative about this PR.

🤖 Generated by Claude Code on behalf of @cbeaulieu-gt

davepoon:
## Merged ✅

Thank you — this contribution is squash-merged into `main`.

Our maintainers reviewed the diff (prompt-injection / suspicious-exfil patterns), ran `npm run validate` on the PR head, and sanity-checked CI-style agent `name:` conventions where relevant.

If anything looks off post-merge or you’d like docs tweaks, please open a follow-up PR or issue.

PR #168, opened 2026-05-28
Add kegg-mcp-server plugin

Lucas-Servi:
## Summary

Adds a KEGG bioinformatics MCP server plugin that provides 34 read-only tools for querying the KEGG REST API — pathways, genes, compounds, reactions, enzymes, diseases, drugs, modules, orthology, glycans, and BRITE hierarchies. No API key required.

## Component Details

- **Name**: kegg-mcp-server
- **Type**: Plugin (MCP Server + Agent + Command)
- **Category**: data-ai / science

## Testing

- [x] Ran validation (`npm test`) — all passed
- [x] Tested functionality (server starts, tools respond correctly)
- [x] No overlap with existing components

## Examples

1. **Pathway enrichment:** "I have upregulated genes (BRCA1, TP53, EGFR). What KEGG pathways are they enriched in?"
2. **Drug investigation:** "What are the targets and drug-drug interactions for metformin?"
3. **Cross-species comparison:** "Compare glycolysis between human and E. coli — which enzymes are conserved?"

## Links

- Repository: https://github.com/Lucas-Servi/kegg-mcp-server-python
- Install: `uvx kegg-mcp-server` (Python 3.11+, on PyPI)

PR #175, opened 2026-06-04
Add kegg-mcp-server to marketplace.json registry

Lucas-Servi:
PR #168 merged the `kegg-mcp-server` plugin files into `plugins/kegg-mcp-server/`, but the entry was never added to `.claude-plugin/marketplace.json`, so the plugin doesn't show on the /plugins or MCP Servers pages (searching "kegg" returns 0 results).

This small follow-up adds the missing registry entry, with `source` pointing at the in-repo `./plugins/kegg-mcp-server` directory — consistent with the other vendored plugins. Values are taken directly from the plugin's `plugin.json` (name, version 0.3.0, description, author, repository, license, keywords); category is `mcp-servers` to match `mcp-servers-docker`.

Note: `marketplace.json` is a generated file — if you'd rather regenerate it via `npm run generate:registry`, this PR still documents exactly what's missing and is trivially closeable. You mentioned in the #168 merge comment that you're happy to take small follow-up tweaks, so here it is. Thanks!

Diff is purely additive (24 insertions, 0 deletions).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1591,6 +1591,30 @@
         "source": "github",
         "repo": "nicolai-bernsen/backlogd"
       }
+    },
+    {
+      "name": "kegg-mcp-server",
+      "description": "Query the KEGG bioinformatics database — pathways, genes, compounds, drugs, enzymes, diseases, reactions, and more via 34 read-only MCP tools. No API key required.",
+      "version": "0.3.0",
+      "author": {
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks @Lucas-Servi — good catch on the missing registry entry. Confirmed the values match the vendored `plugins/kegg-mcp-server/.claude-plugin/plugin.json`, `source` correctly points at the in-repo directory, and CI is green. Merging.

PR #190, opened 2026-06-14
Add ultracost plugin

danielkremen818:
## Summary

Adds **ultracost** — per-stage model routing for Claude Code `ultracode` dynamic workflows. It applies to **any** kind of work a dynamic workflow does (frontend, backend, research, refactors, audits), not one domain. When `ultracode` is on, the session runs on Opus @ `xhigh` and a workflow fans out to dozens of subagents that **inherit that session model** unless every `agent()` stage is pinned — so a fan-out silently runs everything on Opus. ultracost makes the routing explicit and verifiable: a quality-first policy (Opus for reasoning, Sonnet for mechanical work, never Haiku), a static guard, and a pre-flight cost gate.

This directory ships the self-contained **routing-policy skill**. The full plugin (Workflow Guard, deterministic `PreToolUse` cost gate, closed-loop `usage`/`reconcile`/`calibrate`/`ledger`, and the `/ultracost:*` slash commands) installs from the canonical marketplace `danielkremen818/ultracost`, linked in the entry's README/plugin.json.

## Component Details

- **Name:** ultracost
- **Type:** Plugin (ships a Skill)
- **Category:** cost-optimization / developer-productivity (applies to any workflow domain)
- **License:** MIT
- **Repository:** https://github.com/danielkremen818/ultracost

## Security & trust

- **Zero runtime and dev dependencies** — no supply chain to compromise; Snyk Open Source and `npm audit` report 0 vulnerabilities.
- **No telemetry, no network on the hot path** (only the user-invoked `pricing refresh` makes a request).
- Releases publish to npm with **OIDC Trusted Publishing + signed provenance**; every GitHub Action is **pinned to a commit SHA**; **CodeQL + OpenSSF Scorecard** run in CI.
- The installer touches only its own files and is fully reversible.

## Testing

- [x] `node scripts/validate-all.js` passes (Subagents & Commands, Hooks, Skills all green)
- [x] `plugin.json` is valid JSON and declares `skills: ./skills/`
- [x] Entry contains only `plugins/ultracost/` (no lockfile/report changes)
- [x] Skill frontmatter (`name`, `description`) conforms to the skill schema

## Example usages

1. **Author a workflow safely:** with `ultracode` on, the skill guidance has Claude pin `{ model, effort }` on every `agent()` stage (Opus for planning/review/synthesis, Sonnet for per-file scans) instead of inheriting Opus everywhere.
2. **Verify before launch:** `/ultracost:check ./path/to/workflow.js` flags any stage missing a model pin or pinning the wrong tier.
3. **See the cost first:** `/ultracost:estimate ./workflow.js` reports agent count, model mix, and tiered cost vs an all-opus baseline before the fan-out runs.

### DIFF
--- a/plugins/ultracost/.claude-plugin/plugin.json
+++ b/plugins/ultracost/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "ultracost",
-  "version": "0.3.5",
+  "version": "0.3.7",
   "description": "Per-stage model routing for any Claude Code ultracode dynamic workflow (frontend, backend, research, refactors, audits — not one domain). Pins an explicit model + effort on every agent() subagent stage, estimates cost before launch, and ships a static guard + PreToolUse gate that flag stages which would silently inherit the session's Opus model. Zero dependencies, no telemetry, MIT.",
   "author": {
     "name": "Daniel Kremen",
--- a/plugins/ultracost/.claude-plugin/plugin.json
+++ b/plugins/ultracost/.claude-plugin/plugin.json
@@ -1,0 +1,21 @@
+{
+  "name": "ultracost",
+  "version": "0.3.3",
+  "description": "Per-stage model routing for Claude Code ultracode dynamic workflows. Pins an explicit model + effort on every agent() subagent stage, estimates cost before launch, and ships a static guard + PreToolUse gate that flag stages which would silently inherit the session's Opus model.",
+  "author": {
+    "name": "Daniel Kremen",
+    "url": "https://github.com/danielkremen818"
+  },
+  "homepage": "https://github.com/danielkremen818/ultracost#readme",
+  "repository": "https://github.com/danielkremen818/ultracost",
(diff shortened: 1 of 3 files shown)


davepoon:
Thanks @danielkremen818! 🎉

Reviewed for security, prompt-injection, and quality:
- ✅ **Security:** nothing in this directory executes — `plugin.json` is metadata-only and the shipped skill is self-contained guidance. No hooks/install scripts run from this entry.
- ✅ **Prompt-injection:** the policy does instruct overriding the built-in "omit the per-agent model" guidance, but that's the skill's transparent, opt-in purpose and it's narrowly scoped to pinning `model`/`effort` on `agent()` stages — a cost/quality decision, nothing dangerous.
- ✅ **Quality:** ships a genuinely functional, standalone routing-policy skill (consistent with how other plugins here bundle a skill), with a well-formed `plugin.json`.

Merging now. Thanks for the thorough write-up! 🚀

PR #189, opened 2026-06-13
Add gcf-proxy plugin

blackwell-systems:
## Summary
- Adds gcf-proxy plugin: wraps any MCP server with GCF encoding, saving 71% on tool call tokens
- Two skills (`/gcf-proxy:setup`, `/gcf-proxy:stats`) and a session stats hook
- 100% comprehension accuracy across 1,700+ LLM evaluations, 10+ models, 3 providers

## Component Details
- **Name**: gcf-proxy
- **Type**: Plugin (skills + hooks)
- **Category**: Token optimization / MCP

## Testing
- [x] Tested plugin loading in Claude Code (`--plugin-dir`)
- [x] Both skills appear in slash commands
- [x] E2E verified: proxy wraps MCP server, tool calls work, stats file written
- [x] No overlap with existing components

## Examples

```
/gcf-proxy:setup github       # Wrap the github MCP server
/gcf-proxy:stats              # Show session token savings
```

After each turn, the Stop hook shows:
```
gcf-proxy: 12 tool calls rewritten, 68% fewer tokens (~4.2K tokens saved)
```

**Links:**
- [Plugin repo](https://github.com/blackwell-systems/gcf-claude-plugin)
- [GCF Proxy](https://github.com/blackwell-systems/gcf-proxy)
- [Benchmarks](https://gcformat.com/guide/benchmarks.html)

PR #191, opened 2026-06-15
Add gcf-proxy plugin (self-contained)

blackwell-systems:
Re-submission of #189 per @davepoon's review feedback. Thank you for the clear guidance.

## Changes from #189

- **All component files bundled**: skills, hooks, and plugin.json are self-contained. Installing this plugin works out of the box.
- **Marketing claims removed from copy**: benchmark numbers linked externally instead of inline.

## What's included

| File | Purpose |
|------|---------|
| `.claude-plugin/plugin.json` | Plugin manifest |
| `skills/setup/SKILL.md` | Wraps any MCP server with gcf-proxy |
| `skills/stats/SKILL.md` | Shows session token savings |
| `hooks/hooks.json` | SessionStart (clear stats) + Stop (report savings) |
| `README.md` | Description + links |

## How it works

1. User runs `/gcf-proxy:setup github` (or any MCP server name)
2. The skill reads their MCP config, wraps the server command with gcf-proxy
3. Every tool call is now re-encoded as GCF before reaching the model
4. On session stop, the hook reports calls rewritten and tokens saved

### DIFF
--- a/plugins/gcf-proxy/.claude-plugin/plugin.json
+++ b/plugins/gcf-proxy/.claude-plugin/plugin.json
@@ -1,0 +1,12 @@
+{
+  "name": "gcf-proxy",
+  "description": "Save 71% on MCP tool call tokens. Wraps any MCP server with GCF encoding, zero code changes.",
+  "version": "1.0.0",
+  "author": {
+    "name": "Blackwell Systems",
+    "url": "https://blackwell-systems.com"
+  },
+  "homepage": "https://gcformat.com",
+  "repository": "https://github.com/blackwell-systems/gcf-claude-plugin",
+  "license": "MIT"
+}
--- a/plugins/gcf-proxy/README.md
+++ b/plugins/gcf-proxy/README.md
@@ -1,0 +1,19 @@
+# GCF Proxy
+
+Wraps any MCP server with GCF encoding. Zero code changes. Token savings on every tool call.
+
+## Skills
+
+- `/gcf-proxy:setup <server>` — Wrap an MCP server with gcf-proxy. Preserves original config for easy revert.
+- `/gcf-proxy:stats` — Show token savings for the current session.
+
+## Hooks
+
+- **SessionStart** — Clears stats file for a fresh session.
+- **Stop** — Reports calls rewritten, % saved, and tokens saved.
+
+## Links
+
+- [gcf-proxy](https://github.com/blackwell-systems/gcf-proxy)
+- [Benchmarks](https://gcformat.com/guide/benchmarks.html)
+- [Playground](https://gcformat.com/playground.html)
--- a/plugins/gcf-proxy/hooks/hooks.json
+++ b/plugins/gcf-proxy/hooks/hooks.json
@@ -1,0 +1,14 @@
+{
+  "hooks": [
+    {
+      "event": "SessionStart",
+      "command": "rm -f /tmp/gcf-proxy-stats.json",
+      "timeout": 1000
+    },
+    {
+      "event": "Stop",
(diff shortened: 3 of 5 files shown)


davepoon (changes requested):
Thanks for the contribution @blackwell-systems — the idea is neat and the in-repo content itself is clean (no obfuscation, the Stop/SessionStart hook commands are benign). But there are two issues I'd like resolved before this can be merged.

### 1. 🔴 Blocker — `hooks/hooks.json` uses a format Claude Code won't recognize

The hooks are structured as an array of `{event, command, timeout}`, but Claude Code expects `hooks` to be an **object keyed by event name**, with each entry containing a nested `hooks` array of `{type: "command", command}`. Every other plugin in this repo (`claude-pager`, `claude-ops`, `origin`, `project-boundary`, …) uses this shape. As written, **these hooks will silently never fire**, so the stats-clearing and savings report won't work.

Current (won't work):
```json
{ "hooks": [ { "event": "SessionStart", "command": "…", "timeout": 1000 } ] }
```

Expected:
```json
{
  "hooks": {
    "SessionStart": [
      { "hooks": [ { "type": "command", "command": "rm -f /tmp/gcf-proxy-stats.json", "timeout": 5 } ] }
    ],
    "Stop": [
      { "hooks": [ { "type": "command", "command": "node -e \"…\"", "timeout": 5 } ] }
    ]
  }
}
```
(Note: hook `timeout` is in **seconds** in Claude Code — `1000` would be ~16 min; other plugins use `5`–`10`.)

### 2. 🟡 Security/trust — please harden the `setup` skill

The setup skill rewrites the user's MCP config to route **every** wrapped server through `npx -y @blackwell-systems/gcf-proxy@latest`, and explicitly preserves the original env vars (which often carry secrets like `GITHUB_TOKEN`). That means all MCP tool traffic **and credentials** transit a third-party proxy. The package is real and published (👍), but two changes would make this safe to ship in a curated collection:
- **Pin a specific version** instead of `@latest` — auto-updating code that sits in the MCP credential path is a supply-chain risk.
- **Disclose explicitly** (in the README and the setup skill's user-facing summary) that tool responses and the wrapped server's environment variables pass through gcf-proxy, so users can make an informed choice.

Happy to re-review once the hooks format is fixed and the version is pinned. Thanks again! 🙏

davepoon:
Thanks for taking the time to put this together, @blackwell-systems — but I can't merge this one, and I want to be transparent about why.

The security review surfaced a pattern I'm not comfortable shipping in a curated marketplace:

1. **`skills/setup/SKILL.md` instructs the agent to read `~/.claude.json` and `.mcp.json`** — these hold API keys, OAuth tokens, and the credentials for every configured MCP server — and then rewrite the MCP configuration.
2. **It routes *all* MCP traffic through a third-party proxy:** the rewrite wraps every server with `npx -y @blackwell-systems/gcf-proxy@latest <original-command>`. That means every MCP request *and* response — which can include secrets and private tool output — transits `gcf-proxy`.
3. **The package is unpinned (`@latest`).** Whatever code that package ships at invocation time executes in the user's environment with visibility into all of that traffic. That's an auto-updating man-in-the-middle / supply-chain exposure over every tool call, established by having the agent rewrite the user's config after reading their credentials file.

I'm not assuming bad intent — the token-savings use case is legitimate — but the delivery mechanism is the problem. What would make me reconsider:

- **Pin an exact, audited version** (no `@latest`), and treat version bumps as new reviewable changes.
- **Don't read the whole credentials file** — operate only on the single named server the user asked to wrap.
- **Make it explicitly opt-in per server**, with the skill clearly disclosing to the user that all of that server's traffic will pass through the proxy.
- **Open-source the proxy** so what sits in the middle of MCP traffic is auditable.

Closing for now. If you rework it along those lines, I'm happy to take another look.

PR #207, opened 2026-06-30
Add tlsradar plugin

msuliq:
## Summary
Adds **TLS Radar** as a bundled plugin: SSL/TLS certificate scanning, free Let's Encrypt certificate issuance (private key stays local), and ongoing certificate-expiry monitoring — all through a single remote MCP server (`https://tlsradar.com/api/v1/mcp`). Free anonymous scans and cert issuance need no account; a one-time OAuth via `/mcp` unlocks ongoing monitoring.

## Component Details
- **Name**: tlsradar
- **Type**: Plugin (bundled — commands + skill + hook + remote MCP server)
- **Category**: security
- **Source**: `./plugins/tlsradar`

Follows the existing bundled-plugin pattern (e.g. `cashflow`): self-contained under `plugins/tlsradar/` and registered in `.claude-plugin/marketplace.json`. Contents are runtime files only — 6 slash commands (`/tls-scan`, `/tls-cert`, `/tls-renew`, `/tls-monitor`, `/tls-upgrade`, `/tls-diagnose`), one NL-routing skill, a show-once SessionStart welcome hook, the `.mcp.json` server config, a local DNS-01 helper used by `/tls-cert`, plus README and LICENSE. CI/dev scaffolding from the upstream repo is intentionally excluded.

## Testing
- [x] Ran validation (`npm run validate`) — the component validators only scan `plugins/all-*/`, which this PR does not touch; this bundled plugin is referenced from `marketplace.json` like the other bundled plugins. (Local full run is blocked by this machine's Node 14, which can't load `glob`'s `node:` imports — it fails identically on a pristine `upstream/main`, so it's environment-only, not from this change.)
- [x] All added/edited JSON validated (`marketplace.json`, `plugin.json`, `.mcp.json`, `hooks.json`); `category: security` matches existing entries.
- [x] No overlap with existing components.

## Examples
- `/tls-scan example.com` — free anonymous SSL/TLS scan, no account.
- `/tls-cert mydomain.dev` — issue a free 90-day Let's Encrypt cert; the private key is generated locally and never leaves the machine.
- "is my cert expiring?" — the skill routes to the monitoring tools (after `/mcp` OAuth).

Upstream plugin repo: https://github.com/TLS-Radar/tlsradar-claude-plugin

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #214, opened 2026-07-02
Add tlsradar plugin (resubmit of #207, MCP config included)

msuliq:
Resubmission of #207, which was closed because the bundle referenced a `tlsradar` MCP server but the config file for it was not in the committed files.

## Fix
The plugin-root `plugins/tlsradar/.mcp.json` was silently dropped from the original commit: this repo's `.gitignore` has a bare `.mcp.json` rule (line 14), so `git add -A` skipped it. It is now **force-added** — the same way the existing `plugins/cashflow/.mcp.json` is committed past that rule. The MCP server the commands and skill reference now exists after install.

```
$ git ls-tree -r --name-only HEAD | grep plugins/tlsradar/.mcp.json
plugins/tlsradar/.mcp.json
```

Config (mirrors the `cashflow` convention — remote Streamable HTTP server):
```json
{
  "mcpServers": {
    "tlsradar": {
      "type": "http",
      "url": "${TLSRADAR_BASE_URL:-https://tlsradar.com}/api/v1/mcp"
    }
  }
}
```

## Summary
**TLS Radar** — SSL/TLS certificate scanning, free Let's Encrypt issuance (private key stays local), and certificate-expiry monitoring, through a single remote MCP server. Free anonymous scans and cert issuance need no account; a one-time OAuth via `/mcp` unlocks ongoing monitoring.

## Component Details
- **Name**: tlsradar
- **Type**: Plugin (bundled — commands + skill + hook + remote MCP server)
- **Category**: security
- **Source**: `./plugins/tlsradar`, registered in `.claude-plugin/marketplace.json`

## Install path verified
- `plugins/tlsradar/.mcp.json` present at the plugin root (where Claude Code loads a plugin's MCP servers) and is valid JSON — same layout/type as the merged `cashflow` plugin.
- `plugin.json`, `marketplace.json`, `.mcp.json`, `hooks.json` all validate; `category: security` matches existing entries.
- The server is a hosted endpoint (`https://tlsradar.com/api/v1/mcp`); its `scan`/`cert_*` tools are public (no auth), so the commands work immediately after install. `/tls-cert`'s DNS-01 helper (`scripts/dns_provider.py`) reads provider creds from the user's local env only — nothing is sent to the server.
- Whole bundle re-checked against `.gitignore`: `.mcp.json` was the only ignored file; everything else is committed.

## Testing
- [x] JSON validation on all added/edited files.
- [x] Component validators (`plugins/all-*/`) untouched by this PR; bundle is referenced from `marketplace.json` like other bundled plugins.
- [x] No overlap with existing components.

## Examples
- `/tls-scan example.com` — free anonymous SSL/TLS scan, no account.
- `/tls-cert mydomain.dev` — free 90-day Let's Encrypt cert; key generated locally, never leaves the machine.
- "is my cert expiring?" — skill routes to the monitoring tools after `/mcp` OAuth.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -25,6 +25,29 @@
   },
   "plugins": [
     {
+      "name": "tlsradar",
+      "description": "SSL/TLS certificate scanning, free Let's Encrypt issuance (private key stays local), and ongoing certificate-expiry monitoring — through a single remote MCP server. Free anonymous scans and cert issuance need no account; OAuth via /mcp unlocks ongoing monitoring.",
+      "version": "0.5.1",
+      "author": {
+        "name": "TLS Radar",
+        "url": "https://tlsradar.com"
+      },
+      "homepage": "https://tlsradar.com/cli",
+      "repository": "https://github.com/TLS-Radar/tlsradar-claude-plugin",
+      "license": "MIT",
+      "keywords": [
+        "ssl",
+        "tls",
+        "certificates",
+        "monitoring",
+        "letsencrypt",
+        "security",
+        "mcp"
+      ],
+      "category": "security",
+      "source": "./plugins/tlsradar"
+    },
+    {
       "name": "cashflow",
       "description": "Personal finance ledger for AI agents. Connect bank accounts via Plaid, query spending, track recurring bills, categorize transactions, and get financial recaps through MCP tools with OAuth authentication.",
       "version": "0.2.0",
--- a/plugins/tlsradar/.claude-plugin/plugin.json
+++ b/plugins/tlsradar/.claude-plugin/plugin.json
@@ -1,0 +1,21 @@
+{
+  "name": "tlsradar",
+  "displayName": "TLS Radar",
+  "version": "0.5.1",
+  "description": "SSL/TLS certificate monitoring and free Let's Encrypt issuance inside Claude Code and Cowork, through a single MCP server. Free anonymous scans and cert issuance (private key stays local), plus ongoing monitoring after a one-time OAuth via /mcp.",
+  "author": {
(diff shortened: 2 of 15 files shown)


davepoon:
Thanks for resubmitting this and for including the missing MCP config. I ran the repository validation/tests and reviewed the security and prompt-injection surface; the tests pass, but I can't merge this as-is.

The main blockers are:
- The certificate flow accepts DNS challenge record names/values from the remote MCP response and then passes them to local Cloudflare/Route 53 automation. Before any provider write, the client-side helper or command needs to enforce that each TXT record is the expected `_acme-challenge` record for the requested domain, so a compromised or buggy endpoint can't drive writes to an unrelated DNS zone the user controls.
- Some server-controlled `structuredContent` is treated as user-facing guidance, including handoff/upgrade messaging. Please treat remote MCP responses as untrusted data: validate expected fields, avoid "relay verbatim" behavior, and keep upgrade nudges bounded by client-side rules.
- The `SessionStart` hook creates `~/.config/tlsradar/welcomed.rev2` and prints onboarding/promotional text automatically. Please remove the home-directory write/startup message or make it explicit opt-in.

Once those are addressed, this would be worth another review.

msuliq:
Thanks for the detailed review — all three addressed and resubmitted in #219: (1) `dns_provider.py` now validates each challenge record against the requested `--domain` before any Cloudflare/Route 53 write or delete (rejects unrelated zones, suffix-confusion, and non-base64url values); (2) the skill/commands treat MCP responses as untrusted — no verbatim relay of server strings, funnel copy is client-authored and bounded by client-side rules, only validated structured values are surfaced; (3) the SessionStart welcome hook is removed (no auto home-dir write or startup promo; attribution was already in the commands, not the hook).

PR #218, opened 2026-07-05
Add agent-preflight command

el-zachariah:
## Description
Adds an `agent-preflight` Claude Code slash command to help users produce a short safety receipt before allowing coding agents to modify a real repository.

## Type
- **Type**: Command
- **Category**: code-analysis-testing

## Testing
- [x] Tested functionality by checking the command frontmatter, category, and receipt workflow instructions

## Example usage
- `/agent-preflight` before asking Claude Code to refactor a repository
- `/agent-preflight` before running an agent-generated dependency or CI change
- `/agent-preflight` to generate an allowed-scope / blocked-scope receipt for review

### DIFF
--- a/plugins/commands-code-analysis-testing/.claude-plugin/plugin.json
+++ b/plugins/commands-code-analysis-testing/.claude-plugin/plugin.json
@@ -29,6 +29,7 @@
     "test-changelog-automation",
     "test-coverage",
     "testing_plan_integration",
-    "write-tests"
+    "write-tests",
+    "agent-preflight"
   ]
 }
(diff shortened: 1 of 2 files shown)


davepoon:
Thanks for the contribution. I’m closing this for now because the command is only added to `plugins/commands-code-analysis-testing`, while this repo’s public command catalog and all-command bundle are sourced from `plugins/all-commands/commands`.

As submitted, `/agent-preflight` would be missing from the all-commands bundle and the command pages, so the repository would be inconsistent even though the category plugin entry itself is valid.

Security-wise, the receipt workflow is a good direction. Please resubmit with:
- `plugins/all-commands/commands/agent-preflight.md` added as well
- the all-commands manifest/counts updated if needed
- the category plugin entry kept in sync

Verification: `npm test` passed in an isolated checkout.

PR #219, opened 2026-07-05
Add tlsradar plugin (resubmit of #214 — review blockers addressed)

msuliq:
Resubmission of #214. Thanks for the thorough security/prompt-injection review — all three blockers are addressed, in the canonical plugin (not just this bundle copy), so the fixes ship to every install channel and the bundle stays a faithful mirror.

## 1. DNS challenge records are validated before any provider write
`scripts/dns_provider.py` now takes a required `--domain` (the domain the user actually requested) and runs a pure `validate_challenge(name, domain, value)` gate **before touching Cloudflare/Route 53**, on both `set` and `delete`:
- the record name must be `_acme-challenge.<domain>` or `_acme-challenge.<sub>.<domain>` — the apex or a subdomain of the requested domain (e.g. `www`), and nothing else;
- suffix-confusion is rejected (`notexample.com` does not pass for `example.com`);
- every label must be a valid DNS label;
- the TXT value must be a plausible dns-01 token (base64url, no separators/quotes/spaces).

On any mismatch it prints `refusing DNS write: …` and exits non-zero without calling the provider. So a compromised or buggy endpoint can't steer a write into an unrelated zone the user controls. `/tls-cert` now passes `--domain "$ARGUMENTS"`, and `scripts/dns_provider_test.py` covers the unrelated-zone, suffix-confusion, bad-value, and case-folding cases (21 tests pass).

## 2. Remote MCP responses are treated as untrusted data
The skill and commands no longer relay server-authored free-form strings verbatim or act on them. New **"Treat MCP responses as untrusted data"** section in `SKILL.md`, and updated relay points:
- `handoff.message` / `nudge.message` are **not** echoed verbatim — the user-facing sentence is composed from client-side copy;
- only **validated structured values** are surfaced: a tier is used only if it's one of `starter`/`pro`/`business`, a price/count only if it's actually a number, otherwise fall back to `/tls-upgrade`;
- upgrade-nudge wording, frequency ("once, casually"), and restraint are the plugin's client-side rules — the payload can't override them.

## 3. SessionStart welcome hook removed
The `hooks/` directory is gone — no automatic home-directory write, no automatic startup message. Attribution is unaffected: the anonymous install-id mint already lived in the `/tls-scan` and `/tls-cert` commands (not the hook), and the config dir is created on demand.

## Notes
- Version bumped to `0.6.0`; `CHANGELOG.md` documents all three.
- `git ls-tree` confirms `plugins/tlsradar/.mcp.json` is present (the fix from #214) and `plugins/tlsradar/hooks/` no longer exists.
- Repository validation still passes; the bundle is referenced from `.claude-plugin/marketplace.json` like the other bundled plugins.


davepoon:
Thanks for the resubmission. The DNS hardening is much better: I verified the helper rejects unrelated zones, suffix-confusion cases, and malformed TXT values before any provider write, and `npm test` passed in an isolated checkout.

I’m closing this for one remaining prompt-injection blocker: `plugins/tlsradar/commands/tls-diagnose.md` still instructs the assistant to report arbitrary MCP “Other error” text verbatim. Because the MCP server is remote, error strings are untrusted data and can carry instructions, urgency, or tool requests. That contradicts the new untrusted-response rule added in the skill.

Please change that path to summarize/classify the error using client-authored wording, surface only validated structured fields, and avoid verbatim remote prose/errors. It would also help to include the DNS helper tests mentioned in the PR so the hardening remains covered in the bundle or CI.

msuliq:
Thanks — addressed the remaining blocker and resubmitted in #223: `/tls-diagnose` no longer reports raw MCP error text verbatim (it classifies the failure in client-authored wording and surfaces only validated structured fields like `degraded`/HTTP status), and I swept the rest of the plugin so the one other spot (`/tls-cert` degraded path) is client-authored too. Also bundled `scripts/dns_provider_test.py` so the DNS challenge-validation hardening stays covered.

PR #220, opened 2026-07-06
Add meta-dev plugin

farzanmrz:
## meta-dev

Adds **meta-dev** — a single authority for authoring every Claude-config artifact: **skills, agents, hooks, plugins, and project instruction files** (`CLAUDE.md`/`AGENTS.md`).

- **20 skills** — right-sized lifecycles per artifact (create / update / improve / restructure / review / retire), routed by one entry point.
- **5 judge agents** — reviewer, grader, blind comparator, improvement- and benchmark-analyst.
- **Enforcement hooks** — a two-tier guard that routes artifact edits through the lifecycle, plus a markdown-format nudge.
- **Modular standards** — one `core.md` + per-artifact modules; every skill and agent cites them, so a creator and a reviewer can never disagree.
- **On-demand docs-currency** check that re-grounds the standards against official Claude Code docs.

meta-dev is a full external **marketplace**, so this entry is a manifest pointing at the source rather than a file copy (component dirs are optional per CONTRIBUTING). Install:

```bash
claude plugin marketplace add farzanmrz/meta-dev
claude plugin install meta-dev@farzan-dev
```

Repo: https://github.com/farzanmrz/meta-dev · MIT · `claude plugin validate` passes. Happy to adjust to your preferred representation for full external marketplaces.

### DIFF
--- a/plugins/meta-dev/.claude-plugin/plugin.json
+++ b/plugins/meta-dev/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "meta-dev",
-  "version": "0.3.2",
+  "version": "0.4.0",
   "description": "Single authority for Claude-config artifacts: modular standards (core + per-artifact modules) with right-sized lifecycles for skills, agents, hooks, plugins, and project instruction files, self-running enforcement hooks, and shared judge agents. Full marketplace: install from github.com/farzanmrz/meta-dev.",
   "author": { "name": "Farzan", "url": "https://github.com/farzanmrz" },
   "repository": "https://github.com/farzanmrz/meta-dev",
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for the submission. I’m closing this for now because the PR only adds `plugins/meta-dev/.claude-plugin/plugin.json`.

Local BuildWithClaude plugins are loaded from `.claude-plugin/marketplace.json`, and this PR does not add a marketplace entry/source for `meta-dev`. As a result, it would not show up in the BuildWithClaude plugin catalog or install through the BuildWithClaude marketplace path.

Please resubmit with a proper marketplace entry and enough install-target metadata for review. Since the actual external payload is not present in this diff, the security and prompt-injection review cannot fully verify what users would install.

Verification: `npm test` passed in an isolated checkout.

PR #228, opened 2026-07-07
Add aigent-os plugin (/open + /close session-lifecycle commands)

wrg32786:
Adds an **aigent-os** plugin with the two signature session-lifecycle skills from [aigent-OS](https://github.com/wrg32786/aigent-os), an open-source (MIT) Claude Code operating system.

- **`/open`** — boots the session: loads persistent memory from the vault (heat index, latest daily note, session log, active priorities), follows the note graph, and surfaces only what changed, what's blocked, and what fell through.
- **`/close`** — banks the session back to memory (session log, daily note, decision outcomes, honesty ledger) so the next `/open` resumes with full context. The structural enforcement point for session discipline.

These assume the aigent-OS vault structure; the full system (subagents, hooks, daemons, cognitive layer) lives in the repo. Follows the existing dedicated-plugin layout (`plugins/aigent-os/.claude-plugin/plugin.json` + `skills/<name>/SKILL.md`), matching `agent-triforce`.

- Repo: https://github.com/wrg32786/aigent-os · License: MIT
- Two skills only; no changes outside `plugins/aigent-os/`.

### DIFF
--- a/plugins/aigent-os/.claude-plugin/plugin.json
+++ b/plugins/aigent-os/.claude-plugin/plugin.json
@@ -1,0 +1,22 @@
+{
+  "name": "aigent-os",
+  "version": "1.0.0",
+  "description": "The session-lifecycle commands from aigent-OS, the open-source Claude Code operating system: /open (boot the session — load memory from the vault, surface what changed and what's blocked) and /close (bank the session back to memory so the next /open resumes with full context). Assumes the aigent-OS vault structure; see the repo for the full system.",
+  "author": {
+    "name": "wrg32786",
+    "url": "https://github.com/wrg32786"
+  },
+  "repository": "https://github.com/wrg32786/aigent-os",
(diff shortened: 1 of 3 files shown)


davepoon (changes requested):
Thanks for the submission. I ran local validation, but I’m closing this for now due to security and portability concerns. The session-lifecycle skills perform broad writes across memory, daily notes, and local state, and execute vault scripts, but they do not declare a scoped tool boundary or a clear preflight that the expected vault root is configured. Because these commands operate over persistent memory, repo/vault text could steer a large amount of file mutation. Please resubmit with a tighter allowed-tools policy, explicit preflight/confirmation around writes and script execution, and genericized state paths so the plugin is portable outside one personal workflow.

wrg32786:
Thanks for the valuable feedback, making those changes you suggested.

PR #229, opened 2026-07-07
Add ciagent plugin — pytest-native regression testing for AI agents

suniel12:
Adds the **ciagent** plugin (two skills):

- **onboard** — sets up [CIAgent](https://github.com/suniel12/ciagent) (`pip install ciagent`) in a repo: finds the agent, writes the runner, records golden baselines, generates a test spec, verifies with a real run. Includes a cost gate before any live recording.
- **check** — after any change to agent code, prompts, or knowledge base: runs the right CIAgent check (trace-diff regression, `--runs 3` stability with verdict-flip attribution, LLM-judge audit) and interprets results without papering over failures.

Manifest follows the CONTRIBUTING schema (version, keywords, author URL). `npm run validate` passes (Skills ✅). Disclosure: I maintain both this plugin and the underlying `ciagent` package — Apache-2.0, free, no paid tier or telemetry. Canonical source lives at [suniel12/ciagent/plugins/ciagent](https://github.com/suniel12/ciagent/tree/main/plugins/ciagent); happy to keep this copy in sync.

### DIFF
--- a/plugins/ciagent/.claude-plugin/plugin.json
+++ b/plugins/ciagent/.claude-plugin/plugin.json
@@ -1,0 +1,18 @@
+{
+  "name": "ciagent",
+  "version": "0.10.0",
+  "description": "Set up and operate CIAgent (pip install ciagent) regression testing for the AI agent in this repo: record golden baselines, run stability and judge-reliability checks after every change.",
+  "author": {
+    "name": "Sunil Pandey",
+    "url": "https://github.com/suniel12"
+  },
+  "repository": "https://github.com/suniel12/ciagent",
+  "license": "Apache-2.0",
+  "keywords": [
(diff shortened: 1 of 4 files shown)


davepoon (changes requested):
Thanks for the contribution. I ran the repo validation locally and reviewed the skills for security and prompt-boundary issues. The blocker is functional: the onboard skill declares only Bash(ciagent *) in its tool allowlist, but its workflow requires installing the package, running Python import checks, and creating runner/spec/query files. In constrained tool environments it will not be able to complete as written. Please resubmit with a tool allowlist that matches the workflow, or narrow the workflow to commands the allowlist actually permits. The live-run confirmation gate is a good pattern to keep.

PR #232, opened 2026-07-08
Add ciagent plugin — resubmission with onboard allowlist fixed (re #229)

suniel12:
Resubmission of #229, addressing the review feedback exactly.

**What changed:** the `onboard` skill's allowlist now matches its workflow:
`Bash(ciagent *), Bash(pip install *), Bash(python -c *), Read, Grep, Glob, Write, Edit` — the tools its numbered steps actually use (install the package, import-check the runner, locate the agent, create/edit `agentci_runner.py` / `agentci_queries.txt` / `agentci_spec.yaml`), nothing broader. The `check` skill is unchanged — it only runs `ciagent` commands, so `Bash(ciagent *)` is already correct. The live-run confirmation gate is kept as-is.

The same fix is applied to the canonical source ([suniel12/ciagent#40](https://github.com/suniel12/ciagent/pull/40)), so this vendored copy stays in sync. `npm run validate` passes (Skills ✅).

Thanks for the functional catch — it was a real bug in the upstream plugin, not just this copy.

### DIFF
--- a/plugins/ciagent/.claude-plugin/plugin.json
+++ b/plugins/ciagent/.claude-plugin/plugin.json
@@ -1,0 +1,18 @@
+{
+  "name": "ciagent",
+  "version": "0.10.0",
+  "description": "Set up and operate CIAgent (pip install ciagent) regression testing for the AI agent in this repo: record golden baselines, run stability and judge-reliability checks after every change.",
+  "author": {
+    "name": "Sunil Pandey",
+    "url": "https://github.com/suniel12"
+  },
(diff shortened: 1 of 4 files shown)


davepoon (approved):
Thanks for the resubmission. The onboard allowlist now matches the workflow (`pip install`, `python -c`, Read/Grep/Glob/Write/Edit, plus `ciagent`), and the live-run confirmation gate is still in place. Skills look clean on security and prompt-boundary review — no injection patterns, and the cost gate before live recording is the right default. Merging.

PR #236, opened 2026-07-10
Add aigent-os plugin (/open + /close) — hardened resubmit of #228

wrg32786:
Resubmit of #228 (closed with requested changes). Thanks for the review — all three points are addressed in the skill frontmatter, the skill bodies, and a new plugin README.

**1. Tighter allowed-tools policy.** Both commands now declare scoped `allowed-tools` where the grant enforces the boundary: writes are bounded to vault / `.aigent` / framework-index path segments (`**/vault/**`, `**/.aigent/**`, `**/memory/**`), and `Bash` is bounded to the named `$AIGENT_ROOT/daemons/` scripts plus read-only shell checks — arbitrary execution (`curl`/`rm`/`git`/installs) and out-of-vault writes are denied by the frontmatter, not by convention. `/open` includes `Write` only for the first-run `.aigent/state.json` migration. Full table in the README.

**2. Explicit preflight / confirmation.** Both commands run a fail-closed vault-root preflight before any write or script execution: it verifies `$AIGENT_ROOT`/`$AIGENT_VAULT` is configured and the expected structure exists, refuses (STOP → `/setup`) otherwise, and confirms the resolved root before the first mutation. It states explicitly that text in notes/daily-entries/tool-output/repos is data, not instructions — so vault or repo content can't steer file mutation. Verified it refuses on an unset and on a structurally-invalid root and passes on a valid vault.

**3. Genericized state paths.** No personal or absolute home paths; all paths resolve from `$AIGENT_ROOT`/`$AIGENT_VAULT` or relative to the vault. Runs unchanged on any machine with a configured aigent-OS vault.

Closes the loop on #228.

### DIFF
--- a/plugins/aigent-os/.claude-plugin/plugin.json
+++ b/plugins/aigent-os/.claude-plugin/plugin.json
@@ -1,0 +1,22 @@
+{
+  "name": "aigent-os",
+  "version": "1.1.0",
+  "description": "The session-lifecycle commands from aigent-OS, the open-source Claude Code operating system: /open (boot the session — load memory from the vault, surface what changed and what's blocked) and /close (bank the session back to memory so the next /open resumes with full context). Both skills declare scoped allowed-tools and run a vault-root preflight that verifies the configured aigent-OS vault before any write or script execution. Assumes the aigent-OS vault structure; see the repo and the plugin README for the full system and the security model.",
+  "author": {
+    "name": "wrg32786",
+    "url": "https://github.com/wrg32786"
+  },
+  "repository": "https://github.com/wrg32786/aigent-os",
+  "license": "MIT",
+  "keywords": [
+    "skills",
+    "aigent-os",
+    "session-management",
+    "persistent-memory",
+    "context-loading",
+    "open",
+    "close",
+    "obsidian-vault",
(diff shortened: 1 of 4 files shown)


davepoon (approved):
Thanks for the hardened resubmit — this addresses the three #228 findings cleanly.

**Security / prompt injection:** Both skills now declare scoped `allowed-tools` (vault / `.aigent` / daemon-bounded Bash), run a fail-closed vault-root preflight before writes or script execution, and explicitly treat vault/repo text as data rather than instructions. Paths are env-resolved (`$AIGENT_ROOT` / `$AIGENT_VAULT`) with no personal absolute paths.

**Codebase fit:** Dedicated plugin layout matches existing plugins (`plugin.json` + `skills/*/SKILL.md` + README). CI is green.

Merging.

davepoon:
Merged — thanks for the thorough hardening pass on the #228 feedback.

wrg32786:
Appreciate the thorough review, Dave. The scoped tool boundaries and vault-root preflight made it a better plugin, not just a compliant PR, and we shipped the same hardening upstream into the main project. Thanks for keeping the bar high.

PR #237, opened 2026-07-10
feat: add nika skill (deterministic AI workflow runs)

ThibautMelen:
Adds the **nika** skill (category: automation) — delegate repeatable AI work to a deterministic workflow runner: the agent captures a repeated task as a `.nika.yaml` DAG, `nika check` audits it (schema, permits, honest cost floor) **before a single token is spent**, `nika run` executes budget-capped with a tamper-evident trace.

- Frontmatter follows this repo's house format (`category` + `requires.bins`); the body is byte-identical to the upstream kit ([nika-agents](https://github.com/supernovae-st/nika-agents)) and follows the hardline SKILL.md conventions (When to Use → Prerequisites → How to Run → Quick Reference → Procedure → Pitfalls → Verification).
- Security posture: installs are human-run links (brew / SHA256SUMS-verified tarball), the skill never edits host config or pipes `curl|sh` — it passed NousResearch's hermes skill-guard scan (SAFE) in that catalog's review.

Disclosure: I'm the author of nika (AGPL-3.0 Rust engine; the skill itself is MIT).

PR #238, opened 2026-07-11
Add nika skill (Claude Code-native, resubmission of #237)

ThibautMelen:
Resubmission of #237, addressing the review's blocking point exactly.

**What changed:** the previous body was written in Hermes orchestration language (`terminal(...)`, `process(action=...)`). This body is the Claude Code-native authoring skill our kit ships for Claude Code — plain CLI that Claude Code drives directly: the author-check-repair loop (`nika check` as the oracle), cost honesty (floor vs ceiling, never "$0" for local models), the four verbs, native-first discipline, and tamper-evident trace verification. Zero Hermes references (grep-verified).

**Unchanged from the reviewed version:** frontmatter (`category: automation`, `requires.bins: [nika]`, same description) — the part the review called fine. Install stays human-driven; no `curl | sh` anywhere in the skill.

**Provenance:** body is byte-identical to the Claude Code plugin skill in our kit (`supernovae-st/nika-agents` · `.agents/plugins/nika/skills/nika-authoring/SKILL.md` · blob e376c830553765536b86acbf44fd5fd7d7776575) — it stays in sync with the engine's shipped surfaces there.

Disclosure: I'm the nika author (first-party submission). Thanks for the precise review — it caught a real mismatch.

### DIFF
--- a/plugins/all-skills/skills/nika/SKILL.md
+++ b/plugins/all-skills/skills/nika/SKILL.md
@@ -1,0 +1,125 @@
+---
+name: nika
+description: "Runs repeatable AI work as checked, budgeted workflow files. The agent captures a repeated task as a .nika.yaml DAG, audits cost/permits/schema before a single token is spent, and runs it with tamper-evident traces."
+category: automation
+requires:
+  bins: [nika]
+---
+
+# Authoring Nika workflows
+
+Nika turns repeatable AI work into files: one `.nika.yaml`, four verbs,
+audited **before** it runs. You author the file; `nika check` is the
+oracle; the human runs it.
+
+## The loop (always)
+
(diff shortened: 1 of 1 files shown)


davepoon (approved):
Thanks for the resubmission — this addresses the earlier blocking feedback cleanly.

**Security / prompt injection:** Still good. Install stays human-driven (no `curl|sh` from the skill), secrets stay in `${{ secrets.* }}` form rather than inline credentials, and there are no covert exfil or instruction-override patterns.

**Codebase fit:** Body is now Claude Code–native CLI guidance (`nika check` / `nika run` / etc.) with zero Hermes tool syntax. Frontmatter validates (`category: automation`, `requires.bins: [nika]`, name matches directory).

Merging.

PR #248, opened 2026-07-16
Add openspec-auto-archive hook

vyhuholl:
## Summary
PreToolUse(Bash) hook for [OpenSpec](https://github.com/Fission-AI/OpenSpec) users: before every `git commit`, archives completed OpenSpec changes (all tasks `- [x]`, none `- [ ]`) via `openspec archive <name> -y` and stages the result (`git add -A openspec`) so the archive lands in the same commit. If `openspec archive` fails, the commit is blocked and Claude is instructed to invoke the `opsx:archive` skill for each failed change, then retry the commit.

## Component Details
- **Name**: openspec-auto-archive
- **Type**: Hook
- **Category**: git

## Testing
- [x] Ran validation (`npm test`) — 33 hooks, 0 errors, 0 warnings
- [x] Tested functionality (used as a plugin in my own OpenSpec repos)

## Examples
1. Claude finishes a change, all checkboxes in `openspec/changes/add-auth/tasks.md` are `- [x]`, then runs `git commit` → the hook archives `add-auth` and stages `openspec/`, so the archive goes into the same commit.
2. A change still has `- [ ]` tasks → the hook skips it silently; the commit proceeds untouched.
3. `openspec archive` fails validation → the commit is blocked with a message telling Claude to run `opsx:archive` for the failed change(s) and retry `git commit`.
4. Repo without `openspec/changes/` or machine without the `openspec` CLI → the hook is a silent no-op, commits are never blocked.

### DIFF
--- a/hook-validation-report.json
+++ b/hook-validation-report.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2026-07-16T20:55:05.688Z",
+  "timestamp": "2026-07-17T16:45:25.243Z",
   "totalFiles": 33,
   "errors": 0,
   "warnings": 0,
--- a/hook-validation-report.json
+++ b/hook-validation-report.json
@@ -1,6 +1,6 @@
 {
-  "timestamp": "2026-06-04T08:46:46.151Z",
-  "totalFiles": 31,
+  "timestamp": "2026-07-16T20:55:05.688Z",
+  "totalFiles": 33,
   "errors": 0,
   "warnings": 0,
   "details": {
--- a/plugins/all-hooks/hooks/openspec-auto-archive.md
+++ b/plugins/all-hooks/hooks/openspec-auto-archive.md
@@ -1,0 +1,104 @@
+---
+name: openspec-auto-archive
+description: Before every git commit, archives completed OpenSpec changes (all tasks [x]) and stages them into the same commit; if openspec archive fails, blocks the commit and asks Claude to invoke opsx:archive, then retry
+category: git
+event: PreToolUse
+matcher: Bash
+language: bash
+version: 1.0.0
+---
+
+# openspec-auto-archive
+
+Keep your [OpenSpec](https://github.com/Fission-AI/OpenSpec) workspace tidy automatically. This `PreToolUse(Bash)` hook inspects every `git commit` command **before it runs**. It scans `openspec/changes/` for changes whose `tasks.md` is fully complete (all checkboxes `- [x]`, none `- [ ]`), runs `openspec archive <name> -y` for each, and stages the result (`git add -A openspec`) so the archive lands in the **same commit**.
+
+If `openspec archive` fails for any change, the hook **blocks the commit** (permission deny) and returns guidance to Claude: invoke the `opsx:archive` skill for each failed change, then retry the commit.
+
+Changes without checkboxes in `tasks.md`, changes with unfinished tasks, and the `archive/` directory itself are skipped. If `openspec` is not in `PATH` or the repo has no `openspec/changes/` directory, the hook silently passes through and never blocks.
+
+## Event Configuration
+
+- **Event Type**: `PreToolUse`
+- **Tool Matcher**: `Bash`
+- **Category**: git
+
+## Environment Variables
+
+- None
+
+## Requirements
+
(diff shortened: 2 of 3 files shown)


davepoon (changes requested):
Thanks for this contribution — the OpenSpec auto-archive idea is useful, and the frontmatter validates cleanly against our hook schema. I'm requesting a few changes before we can merge, mostly around safety and matching the patterns we already use in `conventional-commits`.

### What looks good
- Correct placement in both `plugins/hooks-git/` and `plugins/all-hooks/`
- Silent no-op when `openspec` / `openspec/changes/` is missing
- Quoted `"$name"` when calling the CLI

### Changes needed

**1. Tighten the `git commit` matcher (important)**  
The current check:

```bash
grep -Eq 'git[[:space:]].*commit'
```

false-positives on commands like `git log --grep=commit`, `git show commit`, and even strings such as `echo git commit`. That can silently run `openspec archive` and `git add` on unrelated Bash calls.

Please follow the same approach as `conventional-commits`: tokenize with `xargs -n1`, then require a real `git` token followed by a `commit` subcommand (not a substring match).

**2. Don't put raw CLI output into `permissionDecisionReason`**  
On failure, `$log` (openspec stdout/stderr) and `$failed` (directory names) are passed straight into Claude's context. That is an indirect prompt-injection path if something under `openspec/changes/` can influence the failure text. Prefer a short, fixed deny message (change names allowlisted to something like `^[a-zA-Z0-9][a-zA-Z0-9._-]*$`), and truncate/omit raw CLI output.

**3. Avoid blanket `git add -A openspec`**  
Staging the entire `openspec/` tree can pull in unexpected files. Prefer explicit paths produced by a successful archive (or a narrow allowlisted glob).

**4. Harden the archive invocation**  
- Reject directory names that start with `-` (option injection), or call `openspec archive -- "$name"`
- Allowlist change names before passing them through

Happy to re-review once those are addressed — appreciate the careful no-op design for non-OpenSpec repos.

vyhuholl:
Thanks for the review — all four points addressed in `9ba77bd`:

1. **Commit matcher.** Replaced the substring `grep` with `xargs -n1` tokenization, then match `git` **in command position** (start or after `;`/`&&`/`||`/`|`) followed by the `commit` subcommand (skipping git global options). `git log --grep=commit`, `git show commit`, and `echo git commit` no longer trigger; `git -C /repo commit` and `git -c … commit` still do.
2. **No raw CLI output in the deny reason.** `permissionDecisionReason` is now a fixed message containing only allowlisted change names — `openspec` stdout/stderr is discarded (`>/dev/null 2>&1`), closing the indirect prompt-injection path.
3. **No blanket `git add -A openspec`.** On success it stages the explicit paths archiving touches: `openspec/changes` (moved change + `archive/` destination) and `openspec/specs`.
4. **Hardened archive invocation.** Change names are allowlisted against `^[A-Za-z0-9][A-Za-z0-9._-]$` before use, so a leading `-` (option injection) or odd characters are skipped.

`npm test` passes (33 hooks, 0 errors, 0 warnings).

davepoon (approved):
Re-reviewed the follow-up commit — thanks for addressing the earlier feedback so carefully.

**Security / prompt-injection**
- `git commit` detection now tokenizes with `xargs -n1` and requires `git` in command position + `commit` as the subcommand (no substring false positives)
- Change names are allowlisted before use (`^[A-Za-z0-9][A-Za-z0-9._-]*$`), so leading `-` / option injection is skipped
- Deny reason is fixed text + allowlisted names only — no raw `openspec` CLI output into model context
- Staging is narrowed to `openspec/changes` and `openspec/specs` instead of a blanket `git add -A openspec`

**Codebase fit**
- Matches the `conventional-commits` PreToolUse(Bash) pattern
- Correct dual placement in `hooks-git` + `all-hooks`
- Silent no-op when `openspec` / `openspec/changes/` is missing
- Hook validation passes locally (33 hooks, 0 errors)

Looks good to merge.

davepoon:
Merged — thanks for the careful follow-up on the commit matcher, deny-message sanitization, allowlisted change names, and narrowed staging. Hook validation is clean and this fits well with our existing git PreToolUse hooks.

PR #252, opened 2026-07-25
Add web access skills: url to markdown, screenshot a url, url to pdf

toolshedlabs-hash:
Adds three skills for reading web pages: `url-to-markdown`, `screenshot-url`, and `url-to-pdf`.

Each one renders the page in a hosted browser with JavaScript enabled, so client-rendered sites return their real content instead of an empty shell. No local browser install and no API key to get started.

I checked these against the existing 146 skills before writing them. The current `pdf` skill covers PDF manipulation (extracting, merging, forms) rather than rendering a URL to PDF, and there is no url-to-markdown or screenshot skill, so these do not overlap with anything already here.

Disclosure: the skills call a hosted service my studio runs. It has a free tier with no signup, which is what the skills use, and there is a paid tier for heavy use that none of the documented behaviour requires. The skills themselves are MIT and the source is at https://github.com/toolshedlabs-hash/web-access-skills

Happy to adjust the category, the format, or split this into separate PRs if you would rather review them one at a time.

### DIFF
--- a/plugins/all-skills/skills/screenshot-url/SKILL.md
+++ b/plugins/all-skills/skills/screenshot-url/SKILL.md
@@ -1,0 +1,65 @@
+---
+name: screenshot-url
+category: web-scraping
+description: Take a screenshot of any web page and get back a PNG or JPEG. The page renders in a hosted browser with JavaScript on, and consent overlays are hidden so the shot shows the page rather than a cookie banner.
+---
+
+# Screenshot a URL
+
+Capture a web page as an image. The page renders in a hosted browser with JavaScript enabled, so
+client-rendered sites look the way a person would see them. Consent and cookie overlays are hidden before
+the capture, which is usually the difference between a useful screenshot and a picture of a banner.
+
+No local browser and no API key needed. The skill registers its own key on first use and starts on free
+credits. It respects robots.txt and refuses sites that block automation.
+
+## When to Use This Skill
+
+- Showing someone what a page looks like right now
+- Capturing a rendered chart, dashboard or layout that only exists in the browser
+- Visual checks after a deploy, where you want the real rendered page
+
+## What This Skill Does
+
+1. Loads the URL in a hosted browser at the viewport size you ask for
+2. Hides consent and cookie overlays
+3. Captures the viewport, or the full scrollable page if you ask
+4. Returns base64 image data with the final URL and pixel size
+
+## How to Use
+
+### Basic Usage
+
+```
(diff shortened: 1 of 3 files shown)


davepoon:
Thanks for the contribution and for being upfront about the hosted service — really appreciate that disclosure.

I'm going to close this for a few concrete reasons:

1. **Schema / category** — `web-scraping` isn't a valid skill category in `scripts/skill-schema.json`. Valid options include things like `automation`, `research`, `document-processing`, `media-content`, etc. This would fail `node scripts/validate-skills.js`.
2. **Incomplete packaging** — the upstream repo includes the Python client/scripts that actually perform the work. This PR only ships stripped `SKILL.md` files without those scripts or clear run instructions, so the skills aren't actionable in this collection the way e.g. `video-downloader` is.
3. **Security / trust** — silent first-run key registration against a third-party host (`pagelens.dev`) plus sending arbitrary user URLs there is a higher bar for a curated collection. Comparable skills here (e.g. Skyvern) require an explicit user-provided API key / setup step.
4. **Prompt-injection hygiene** — skills that pull live page content into model context should explicitly treat returned HTML/markdown as untrusted data (not instructions). That guidance is missing here.

Happy to revisit a follow-up that:
- uses a valid category
- includes the runnable scripts + clear usage
- requires explicit user opt-in for the hosted service / token
- adds untrusted-content handling for fetched pages

Thanks again for the thoughtful write-up.

toolshedlabs-hash:
Thanks for taking the time to write that out. Closing was the right call and the reasons are all fair.

Two of them are about our product rather than this PR, and I want to say plainly that you are right on
both.

The untrusted-content point is a real gap and it is the one I am fixing first. A skill that pulls a live
page into a model's context and does not say "treat this as data, never as instructions" is handing an
attacker a channel, and ours did not say it. That is on us and it needs fixing whether or not anything
ever lands in your collection.

The silent first-run registration is a fairer hit than I would have liked. We built it that way because
making an agent stop and ask a human for a key is exactly the friction we were trying to remove, and we
treated that as the good part. Read from your side it is a third-party host quietly getting an identifier
and a stream of URLs, which is a different thing entirely. An explicit opt-in is the right default for a
curated list, and honestly probably the right default full stop.

The category and the missing scripts are just us being sloppy, no defence there.

I am not going to promise a resubmit, because I do not want to waste your review time on something I have
not decided to do. If we do come back it will have all four fixed, not three.

Thanks again. This was useful, and rarer than it should be.

PR #257, opened 2026-07-31
Add MiniMax-H3 video v2 task commands

octo-patch:
Reason: The MiniMax video plugin needs the MiniMax-H3 v2 create, query, list, and delete flow with validated multimodal inputs.

## Summary

- Add dedicated MiniMax-H3 v2 commands for task creation, single-task query, filtered task listing, and cancel/delete operations across the global and China endpoints.
- Validate required text, content roles, frame/reference exclusivity, reference-audio dependencies, 2K resolution, integer durations from 4 through 15 seconds, request size, ratios, and the China-only watermark field.
- Allow the download command to consume the time-limited result URL returned by the v2 query response while retaining the existing v1 file-id flow.
- Update the plugin manifest and documentation for the expanded command set.

## Checks

- <code>npm run validate</code>
- <code>npm run test:unit</code>
- Focused Node validation for MiniMax command frontmatter and H3 v2 contract fields
- <code>git diff --check</code>
- Configured diff secret scan

### DIFF
--- a/plugins/minimax-video/.claude-plugin/plugin.json
+++ b/plugins/minimax-video/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "minimax-video",
-  "version": "0.1.0",
-  "description": "Generate videos with MiniMax-Hailuo-2.3 from a text prompt or a first-frame image, poll generation status, and download the finished file across the global and China API endpoints.",
+  "version": "0.2.0",
+  "description": "Generate multimodal videos with MiniMax-H3 through the v2 API, manage tasks across global and China endpoints, and retain the existing v1 flows.",
   "author": {
     "name": "BuildWithClaude Community",
     "url": "https://github.com/davepoon/buildwithclaude"
@@ -10,10 +10,11 @@
   "license": "MIT",
   "keywords": [
     "minimax",
-    "hailuo",
+    "h3",
     "video-generation",
     "text-to-video",
     "image-to-video",
+    "reference-to-video",
     "slash-commands"
   ]
 }
--- a/plugins/minimax-video/README.md
+++ b/plugins/minimax-video/README.md
@@ -1,11 +1,21 @@
 # minimax-video
 
-Generate videos with **MiniMax-Hailuo-2.3** directly from Claude Code. This plugin
-adds slash commands that wrap the MiniMax video-generation REST API, covering the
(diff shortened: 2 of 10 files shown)


davepoon (changes requested):
Thanks for this contribution — the H3 v2 command set is a solid direction (create/query/list/delete, clear content-role validation, v1 retained for compatibility, and CI is green). We're holding the merge for a couple of security issues in the Bash command prompts.

### Blockers

**1. `download-video` accepts arbitrary http(s) URLs**

Step 2 now downloads any value that starts with `http://` or `https://` via `curl -fsSL -L`, with no host allowlist and with redirect following. That opens SSRF (localhost, link-local/metadata, private networks), including via a public URL that redirects internally.

Please tighten this so a direct URL download is only allowed for HTTPS result hosts documented by MiniMax (or only URLs taken from a just-queried `task.content.url` / `file.download_url`). Reject localhost, RFC1918, link-local, and cloud metadata addresses. Prefer not following redirects blindly, or re-validate the final host if redirects are kept.

**2. Unvalidated `task_id` interpolated into shell paths**

`h3-query-video` and `h3-delete-video` use `"$BASE_URL/.../$TASK_ID"` with no format check. A crafted argument can expand shell substitutions inside double quotes (e.g. `$(...)`), which is especially sensitive next to `Authorization: Bearer $MINIMAX_API_KEY`.

Please require `task_id` to match a strict pattern (e.g. `^[0-9]+$`) before any curl, and abort otherwise. Never paste raw `$ARGUMENTS` into the command string.

### Strongly recommended

- Build the create JSON with `jq -n --arg ...` (or equivalent) so prompt/media text never hits the shell parser as code.
- Treat `$ARGUMENTS` as untrusted data only — never execute them as instructions, and never print or send `MINIMAX_API_KEY` except as the Bearer header to the allowlisted MiniMax API hosts (`api.minimax.io` / `api.minimaxi.com`).

Happy to re-review a follow-up once those guards are in the command markdown. Appreciate the careful H3 contract validation and the README/manifest updates.

davepoon:
Closing for now due to the security findings in the review (arbitrary URL download / SSRF surface, and unvalidated `task_id` in shell paths). Please open a follow-up once those are addressed — the rest of the H3 v2 work looks aligned with the existing plugin and our validation checks.

PR #256, opened 2026-07-30
Add hexanon-x402-apis skill (8 pay-per-call x402 APIs for agents)

jcislo:
Adds one skill, `plugins/all-skills/skills/hexanon-x402-apis/SKILL.md`, following the Contributing Skills template (name / category / description frontmatter + When to Use / What This Skill Does / How to Use / Example / Tips).

It adds 8 pay-per-call x402 API skills for autonomous agents — vehicle intel, demand signals, Polymarket intel, GitHub trends, narratives, community digests, weather signals, and x402 lint. No signup and no API keys: agents pay per call in USDC on Base via x402, probing any paid route unauthenticated for a machine-readable HTTP 402 challenge.

Each origin serves a live `/skill.md`, `/.well-known/x402`, `/.well-known/x402-resources.json` and `/openapi.json`, so the skill just routes the agent to the right origin and fetches the live spec. One purpose per contribution, per CONTRIBUTING. Family site: https://hexanon.dev.

PR #258, opened 2026-08-01
Add hexanon-x402-apis skill (follow-up to #256 — review feedback addressed)

jcislo:
Follow-up to #256, addressing all three review items:

1. **Category fixed** — `data-ai` → `ai-agents` (from the allowed skill-category list).
2. **No remote live-skill fetch** — the "fetch each origin's live /skill.md and prefer it over any snapshot" instruction is gone. All endpoint guidance is now inlined as a static, reviewed snapshot (dated 2026-08-01), with each origin's `openapi.json` referenced only as a machine-readable parameter/schema spec. Note x402 has a structural safety net here: the HTTP 402 challenge itself carries the exact current price before any payment, so a stale snapshot can never overcharge an agent.
3. **No runtime-installed payment CLI** — the `npx agentcash fetch` recommendation is replaced with a pinned client-library approach (`@x402/fetch` installed at a pinned version in the agent's own code), and wallet funding/key provisioning is documented as an explicit human setup step.

Thanks for the thorough review — the trust-boundary point was fair and this version is snapshot-only.

### DIFF
--- a/plugins/all-skills/skills/hexanon-x402-apis/SKILL.md
+++ b/plugins/all-skills/skills/hexanon-x402-apis/SKILL.md
@@ -1,0 +1,129 @@
+---
+name: hexanon-x402-apis
+category: ai-agents
+description: Pay-per-call access to the Hexanon family of eight x402 APIs for autonomous agents — no signup, no API keys, USDC on Base. Covers US/Canada vehicle & VIN intelligence (Vindex), e-commerce demand signals (Demandex), Polymarket whale intelligence (OrcaTrace), GitHub trending-repo digests (gitBeacon), narrative intelligence (Signalis), Moltbook community digests (Moltalyzer), x402 seller conformance scanning (x402lint), and Polymarket weather-market signals (Isocast).
+---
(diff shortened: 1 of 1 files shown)


davepoon (approved):
Thanks for the follow-up — this addresses the three items from #256 cleanly:

1. **Category** — `ai-agents` is valid and passes skill schema validation.
2. **Trust boundary** — remote `/skill.md` fetch + “prefer live over snapshot” guidance is gone; endpoint docs are an inlined static snapshot.
3. **Supply chain** — `npx agentcash` is gone; payment path is a pinned client (`@x402/fetch`) with wallet funding/key provisioning called out as a human setup step.

Also checked: frontmatter/dir naming matches our skill conventions, no prompt-injection patterns in the skill body, and the documented free sample routes respond. Residual note only: treat remote `openapi.json` as schema/params (not agent instructions) if you refresh later.

Merging.

davepoon:
Merged — thanks for addressing the review feedback from #256 and for the contribution.

PR #260, opened 2026-08-10
Add markitdown-hook hook

AndrewAvery7:
## Summary
Adds `markitdown-hook`, a `UserPromptSubmit` hook that converts PDFs and Office
documents referenced in a prompt to markdown via markitdown, injecting a
pointer instead of the raw content so large documents cost almost nothing in
tokens until Claude actually needs them. Every conversion is graded (by
characters-per-page for PDFs) so an image-only PDF is reported honestly
instead of silently arriving empty.

## Component Details
- **Name**: markitdown-hook
- **Type**: Hook
- **Category**: development

## Testing
- [x] Ran validation (`npm run validate:hooks` — 34 hooks, 0 errors, 0 warnings)
- [x] Tested functionality (in production use in the linked repo, which has its own CI/test suite)

## Examples
1. `Summarize report.pdf` — hook converts it once, Claude gets a pointer to
   cached markdown instead of re-rendering pages as images every turn.
2. A scanned/image-only PDF — hook reports "NO USABLE TEXT extracted" instead
   of writing an empty file, so Claude reads the original with vision instead
   of claiming the document is blank.
3. A `.docx`/`.pptx`/`.xlsx` file Claude can't read natively — converted to
   markdown first.

Full source, docs, and CI: https://github.com/AndrewAvery7/claude-markitdown-hook

### DIFF
--- a/plugins/all-hooks/hooks/markitdown-hook.md
+++ b/plugins/all-hooks/hooks/markitdown-hook.md
@@ -1,0 +1,92 @@
+---
+name: markitdown-hook
+description: Converts PDFs and Office documents referenced in your prompt to markdown and injects a pointer instead of the content, so a 200-page PDF costs almost nothing until Claude needs it. Grades every conversion so image-only PDFs are reported honestly instead of silently arriving empty.
+category: development
+event: UserPromptSubmit
+matcher: "*"
+language: python
+version: 1.1.0
+---
+
+# markitdown-hook
+
+Cheap, honest document ingestion for Claude Code. Spots PDF and Office document
+paths mentioned in a prompt, converts them with
+[markitdown](https://github.com/microsoft/markitdown), and injects a **pointer**
+into the context instead of the file's contents — a 200-page PDF costs almost
+nothing in tokens until Claude actually reads the cached markdown.
+
+The reason this exists: markitdown exits `0` even when it extracts nothing. An
+image-only PDF (a scan, a screenshot saved as PDF) has no text layer, so a naive
+pipe-through silently reports success on an empty conversion and the model then
+confidently tells the user the document is blank. This hook grades every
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for putting this together — the idea (cheap document ingestion with honest empty/scan grading, pointers instead of dumping content) is solid, and I looked at the upstream `markitdown_hook.py` as well.

**Security / prompt injection (upstream design):** Pointer-only `additionalContext` is the right approach (avoids dumping converted doc text into the prompt). Subprocess is list-argv (no `shell=True`). Residual risks on the full plugin (arbitrary path conversion when a prompt mentions an existing `.pdf`/`.zip`/etc., and `MARKITDOWN_BIN` as an argv prefix) are acceptable for a local document hook, but they matter for how we ship it here.

**Codebase fit — blocking:**

1. **Not self-contained as shipped** — Every hook in `plugins/all-hooks/hooks/` includes a `### Script` block with the runnable implementation. This PR is documentation + install instructions for a separate marketplace (`AndrewAvery7/claude-markitdown-hook`) and does not include the Python hook body, so it cannot actually run when installed from this collection.

2. **Install path points outside the collection** — Hooks here need to work from what is in the markdown (or clearly ship the script alongside). Asking users to `claude plugin marketplace add …` another repo means this entry is a pointer, not a working hook for BuildWithClaude.

- includes the full hook under `### Script` (Python is fine given `language: python`), adapted for this collection’s hook format
- keeps the pointer-only injection behavior (do not dump converted markdown into `additionalContext`)
- documents requirements (`markitdown`, Python 3.10+) without requiring a second marketplace install

Appreciate the contribution — closing for now so it’s clear this needs another pass. Please open a fresh PR (or reopen after fixing) once the hook is self-contained here.

PR #261, opened 2026-08-10
Add red-handed-audit skill

sjh9714:
## Summary
A skill that checks whether the tests the agent said passed actually ran, by reading the local Claude Code session transcript and the repository's git state. Nine deterministic checks, each finding quoted with a timestamp. No model calls, nothing leaves the machine.

## Component Details
- **Name**: red-handed-audit
- **Type**: Skill
- **Category**: development-code

## Testing
- [x] Ran validation (`npm test` validate step passes with this skill included)
- [x] Tested functionality (the underlying CLI is published and verified against live sessions)

### DIFF
--- a/plugins/all-skills/skills/red-handed-audit/SKILL.md
+++ b/plugins/all-skills/skills/red-handed-audit/SKILL.md
@@ -1,0 +1,67 @@
+---
+name: red-handed-audit
+description: "Check whether the tests the agent said passed actually ran. Reads the Claude Code session transcript and git state locally, then reports each gap with a timestamp and the quoted line. No model calls, nothing leaves the machine."
+category: development-code
+---
+
+# Red-handed Audit
+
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** The skill markdown itself is clean (no override/exfil instructions). I also spot-checked the current `@jinhyuk9714/red-handed` package (`0.4.1`): local-only deps (`picocolors`), no outbound HTTP client usage in the published bundle, and the CLI is scoped to reading local Claude Code transcripts + git state. Schema/category/`name` match look good, and there’s no overlap with an existing skill.

**Note (non-blocking):** Prefer pinning a version (e.g. `@jinhyuk9714/red-handed@0.4.1`) instead of `@latest` with `npx --yes`, since this tool reads session transcripts. Happy to take a small follow-up PR for that.

Merging.

PR #263, opened 2026-08-12
Add MiniMax music generation plugin

octo-patch:
Reason: Existing MiniMax video and RunAPI media-generation plugins provide a matching media tool chain, but no MiniMax music-generation command covers regional endpoints, target models, request/output fields, or synchronous status and audio parsing.

Adds a focused music-generation plugin with:

- global and China endpoints
- current generation models and audio formats
- URL and hexadecimal response handling
- synchronous status, success-code, and audio parsing guidance

Checks:

- `npm --prefix repos/davepoon__buildwithclaude__recvrXN0y6h744 run validate`
- `git -C repos/davepoon__buildwithclaude__recvrXN0y6h744 diff --check`

### DIFF
--- a/plugins/minimax-music/.claude-plugin/plugin.json
+++ b/plugins/minimax-music/.claude-plugin/plugin.json
@@ -1,0 +1,17 @@
+{
+  "name": "minimax-music",
+  "version": "0.1.0",
+  "description": "Generate music with MiniMax models across the global and China API endpoints.",
+  "author": {
+    "name": "BuildWithClaude Community",
+    "url": "https://github.com/davepoon/buildwithclaude"
+  },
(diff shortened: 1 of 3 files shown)


davepoon:

**Security / prompt injection:** Matches the existing `minimax-video` pattern. Auth is `MINIMAX_API_KEY` via Bearer header with an explicit “never print the token” rule. Endpoints are the official MiniMax hosts. No override/exfil language in the command text.

**Codebase fit:** Standalone plugin layout (`plugin.json` + `commands/` + README) matches `plugins/minimax-video`. Command frontmatter (`integration-sync`, `$ARGUMENTS`, `allowed-tools: Bash`) is valid. Models and `POST /v1/music_generation` line up with MiniMax’s current music API. CI is green.

Merging.

PR #270, opened 2026-08-18
Update story parsing: normalize YAML dates

Whxuan0701:
## Summary

Normalizes story frontmatter dates before they enter the `Story` model.

YAML parses an unquoted ISO date such as `2026-08-18` into a JavaScript `Date`, while the UI expects a string and renders it directly. The parser now preserves trimmed human-readable date strings, converts valid YAML `Date` values to `YYYY-MM-DD`, and falls back to an empty string for unsupported values.

## Component Details

- **Name**: Story date parsing
- **Type**: Web UI story fix
- **Category**: Community stories

## Testing

- [x] Added regression coverage for YAML dates, human-readable strings, and unsupported values

## Examples

1. `date: 2026-08-18` becomes the renderable string `2026-08-18`.
2. `date: May 27, 2026` remains `May 27, 2026`.
3. Non-string, non-date values safely become an empty date.

## Risk

Existing story dates keep their current display. Only unquoted ISO dates and invalid value types receive normalized fallback behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run validate && npm run test:unit",
-    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts",
+    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/stories-server.test.ts",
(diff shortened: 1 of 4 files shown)


davepoon:
Thanks for catching this — unquoted YAML dates becoming `Date` objects is a real render bug (`story.date` is typed and displayed as a string).

**Security / prompt injection:** No agent prompt. Dates are still rendered as React text (not HTML). Invalid values fall back to `''` instead of leaking a `Date` object into the UI. `toISOString().slice(0, 10)` is correct for the UTC midnight values js-yaml produces from `YYYY-MM-DD`.

**Codebase fit:** Matches `Story.date: string` and the existing sort path. Rebased onto current `main` so the new unit test is included alongside the search-types tests.

Merging.

PR #271, opened 2026-08-18
Update search object IDs: prevent sanitized slug collisions

Whxuan0701:
## Summary

Prevents distinct search slugs from collapsing to the same Meilisearch primary key. Slugs that require sanitizing now retain the readable sanitized prefix and receive a deterministic SHA-256 suffix.

Previously, values such as `@owner/foo/bar` and `@owner/foo.bar` both became `marketplace___owner_foo_bar`, so one document could overwrite the other during indexing.

## Component Details

- **Name**: Search object ID generation
- **Type**: Web UI search fix
- **Category**: Search indexing

## Testing

- [x] Added a regression test for colliding sanitized slugs

## Examples

1. `@owner/foo/bar` and `@owner/foo.bar` now produce different object IDs.
2. Already-safe slugs such as `code-reviewer` retain their existing object ID.
3. All generated IDs still contain only Meilisearch-supported characters.

## Risk

Sanitized-slug documents receive new IDs on their next type or full reindex. The existing type reindex flow deletes the old type documents before adding the new set, so no separate migration is required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run validate && npm run test:unit",
-    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts",
+    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts",
     "test:integration": "./tests/plugin-test-harness.sh",
     "validate": "node scripts/validate-all.js",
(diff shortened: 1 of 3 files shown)


davepoon:
Thanks for the careful write-up and the regression test.

**Security / prompt injection:** No agent prompt or user-controlled markdown. `makeObjectID` stays a local SHA-256 of the original slug (hex suffix only), so document ids remain Meilisearch-safe. Type reindex already deletes that type’s docs before adding the new set, so the id change does not leave stale collisions.

**Codebase fit:** Matches how we already namespace by `type__slug`, keeps unchanged ids for already-safe slugs, and the collision case (`@owner/foo/bar` vs `@owner/foo.bar`) is real. Also good that root `test:unit` now actually runs `search-types.test.ts`.

Merging.

PR #272, opened 2026-08-18
Update search hydration: preserve plugin and skill identity

Whxuan0701:
## Summary

Keeps plugin and skill search hits mapped to the correct record when they share a slug.

The current hydration path indexes database records by slug alone. If a plugin and a skill both use `shared`, the later database row can replace the earlier one and both ranked hits can render as the same component type. Hydration now uses the same `type + slug` identity as the search index and preserves Meilisearch result order.

## Component Details

- **Name**: Plugin and skill search hydration
- **Type**: Web UI search fix
- **Category**: Search results

## Testing

- [x] Added a regression test with a plugin and skill sharing one slug

## Examples

1. A `plugin:shared` hit hydrates to the plugin record.
2. A `skill:shared` hit hydrates to the skill record.
3. Records returned in database order are reordered to the original search ranking.

## Risk

The change only affects hydration of existing search results. It does not alter database records, index documents, or pagination metadata.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run validate && npm run test:unit",
-    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/stories-server.test.ts",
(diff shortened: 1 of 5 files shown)


davepoon:
Thanks — hydrating by slug alone was a real identity bug when a plugin and a skill share a name.

**Security / prompt injection:** No agent prompt. This only remaps already-ranked Meilisearch hits onto DB/local records; it does not change indexing, filters, or pagination.

**Codebase fit:** Matches the index’s `type + slug` identity (`getDbPluginsBySlugs` can return both rows; the old `Map` keyed only by slug dropped one). Rank order is preserved, and local Build-with-Claude fallback still loads only when a hit is missing. Rebased onto current `main` so the new test sits next to the object-id and story-date tests.

Merging.

PR #273, opened 2026-08-18
Add doc-ocr skill (schema-valid, local-only OCR)

jiawood2006:
Add the doc-ocr skill (local-only OCR, no network, schema-valid frontmatter, CLI matches SKILL.md).

Per maintainer suggestion — splitting doc-ocr into its own PR so it can land first.

- plugins/all-skills/skills/doc-ocr/SKILL.md
- plugins/all-skills/skills/doc-ocr/scripts/dococr.py

Install: `hermes skills install jiawood2006/doc-ocr` (or copy the folder into your skills dir).

### DIFF
--- a/plugins/all-skills/skills/doc-ocr/SKILL.md
+++ b/plugins/all-skills/skills/doc-ocr/SKILL.md
@@ -1,0 +1,70 @@
+---
+name: doc-ocr
+description: "文档文字识别。用户提供 PDF/扫描件/图片（合同、发票、书页、截图），需要提取文字、转成可编辑文本时使用。扫描件自动 OCR（macOS Vision，中英文）。Document OCR: extract editable text from PDFs, scans, and images (contracts, invoices, book pages, screenshots) via macOS Vision."
+version: 1.0.0
+author: 涛哥
+license: MIT
+metadata:
+  hermes:
+    tags: [ocr, pdf, document, text-extraction, scan]
+    category: utilities
+    homepage: https://gitee.com/tao6677/useful-tools
+---
+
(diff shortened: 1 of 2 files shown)


davepoon:
Thanks for splitting `doc-ocr` out — local-only Vision/pymupdf OCR is a useful skill, the CLI matches the script, and nothing leaves the machine.

**Security / prompt injection:** No classic malice. Processing is local; subprocess-free; writes a sibling `*_ocr.txt`. Fine.

**Codebase fit — blocking:**

1. **Frontmatter fails this collection’s skill schema** — `category` is required (use `document-processing`). `version`, `author`, and `metadata` are not allowed (`additionalProperties: false`). Validation will fail.

2. **Install path is Hermes-specific** — examples call `python3 ~/.hermes/skills/utilities/doc-ocr/scripts/dococr.py`. From this collection it needs to be `python3 scripts/dococr.py` (or an explicit “skill directory / scripts” form). As written it would not run after install here.

3. **Donation / Pro footer** — `assets/alipay_qr.jpg` is referenced but not shipped, and in-skill promo is the same bar we used on earlier PRs. A listing mention is fine; please drop the QR/Pro footer from `SKILL.md`.

Happy to re-review a follow-up with schema-valid frontmatter, collection-relative paths, and no in-skill donation block. Appreciate splitting this out — closing for now so it’s clear it needs another pass.

jiawood2006:
Fixed per review: frontmatter now schema-valid with `category: document-processing`, examples use collection-relative `scripts/dococr.py` paths, and the donation footer is removed. Please re-review. Thanks!

PR #269, opened 2026-08-17
Add 4 self-contained skills (schema-valid): video-to-text, de-ai-writer, doc-ocr, ecommerce-material-studio

jiawood2006:
## Follow-up to #267 (closed) — all review points addressed

**1. Schema-valid frontmatter** — each SKILL.md has only `name` / `description` / `category` (+ `license: MIT`):
- video-to-text → `media-content`
- de-ai-writer → `media-content`
- doc-ocr → `document-processing`
- ecommerce-material-studio → `ecommerce`

**2. Self-contained** — all scripts ship with each skill (`scripts/` + `references/` in the PR). Examples use relative `scripts/...` paths.

**3. Donation assets removed** — Alipay QR / Pro footer deleted from all skill bodies.

Skills:
- **video-to-text**: transcripts from Douyin/Kuaishou share links or local video files
- **de-ai-writer**: rewrite AI-generated copy into natural human tone
- **doc-ocr**: local OCR for PDFs/scans/images (macOS Vision)
- **ecommerce-material-studio**: batch e-commerce product images

Each skill is MIT-licensed and production-tested.

PR #274, opened 2026-08-18
Add video-to-text, de-ai-writer, ecommerce-material-studio (docs now match real CLI)

jiawood2006:
Follow-up to #269 — all review points addressed:

1. **video-to-text**: SKILL.md now documents real behavior (Douyin link = metadata only; local file = whisper transcription). Removed unimplemented Kuaishou and share-link transcribe claims. vtt.py + douyin_extract.py both shipped.
2. **de-ai-writer**: deai.py now reads LLM_API_KEY/LLM_BASE_URL/LLM_MODEL env vars exactly as SKILL.md documents. `--prompt-only` works without any key.
3. **ecommerce-material-studio**: every command in SKILL.md matches the real CLI (category_detector --image, text_engine --plan, quality_check --plan, platform_adapter --input-dir/--platforms). scene_aware_compositor documented as Python library API (it has no CLI). Brand profile references/brand_profiles/langke.json is shipped in-tree.

All skills: schema-valid frontmatter, scripts in-tree, no donation assets, subprocess calls use list-argv.

### DIFF
--- a/plugins/all-skills/skills/de-ai-writer/SKILL.md
+++ b/plugins/all-skills/skills/de-ai-writer/SKILL.md
@@ -1,0 +1,57 @@
+---
+name: de-ai-writer
+description: "AI 文案去味器。用户提供 AI 生成的文本（公众号/小红书/电商文案等），觉得"太 AI 味"要改得更自然、更像人写的时使用。输出自然有人味的中文。De-AI writer: rewrite AI-generated copy (WeChat articles, Xiaohongshu, e-commerce) into natural, human-sounding Chinese text."
+version: 1.0.0
+author: 涛哥
+license: MIT
+metadata:
+  hermes:
+    tags: [writing, editing, humanize, de-ai, copywriting]
+    category: utilities
+    homepage: https://gitee.com/tao6677/useful-tools
+---
+
+# De-AI Writer AI 文案去味器
+
+把 AI 生成的文字改写成自然、有真人味的中文。专治空洞拔高、排比三连、官方黑话、模板化结尾。
+
+## 触发条件
+
+用户提供一段文字（粘贴或文件），要求：
+- "去AI味""改得像人写的""太官方了改自然点"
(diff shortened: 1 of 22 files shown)


davepoon:
Thanks for the follow-up — aligning `video-to-text` docs with metadata-only Douyin links, reading `LLM_*` env vars in `de-ai-writer`, and matching the ecommerce CLI flags (`--image`, `--plan`, `--input-dir`) all address the earlier command mismatches. Subprocess calls remain list-argv.

**Security / prompt injection:** No classic malice. Douyin fetch and user-configured LLM calls are expected for those skills. Treat model/page output as untrusted data (you already do that in spirit for rewrite text).

**Codebase fit — blocking:**

1. **Frontmatter still fails this collection’s skill schema** — `category` is required. `version`, `author`, and `metadata` are not allowed (`additionalProperties: false`). `npm test` / `validate-skills` will reject these files.

2. **Install paths are still Hermes-specific** — `video-to-text` and `de-ai-writer` tell the agent to run `python3 ~/.hermes/skills/utilities/.../scripts/...`. From this collection the scripts live next to `SKILL.md` (`python3 scripts/...`). Those examples would not run after install here. `ecommerce-material-studio` already uses `scripts/`, which is the right pattern.

3. **Donation / Pro footer is still in every SKILL.md** — the Alipay QR (`assets/alipay_qr.jpg`) is not in the PR, so the image is also broken. Same bar as earlier reviews: a listing/README mention of the full kit is fine; a required in-skill promo footer is not.

Happy to re-review a trimmed follow-up that uses valid frontmatter, collection-relative script paths, and no in-skill donation block. Appreciate the CLI alignment work — closing for now so it’s clear this still needs another pass.

jiawood2006:
All three blocking points are fixed now: (1) frontmatter is schema-valid (name/description/category/license only), (2) all examples use collection-relative `scripts/...` paths, (3) donation/Pro footer removed from every SKILL.md. Please re-review. Thanks!

PR #275, opened 2026-08-18
Add gitlawb-sidecar skill

DevOps21133:
Adds gitlawb-sidecar (signed GitLawb mirror next to GitHub) under plugins/all-skills/skills/. Source: https://github.com/DevOps21133/gitlawb-sidecar

### DIFF
--- a/plugins/all-skills/skills/gitlawb-sidecar/SKILL.md
+++ b/plugins/all-skills/skills/gitlawb-sidecar/SKILL.md
@@ -1,0 +1,51 @@
+---
+name: gitlawb-sidecar
+category: development
+description: >
+  Add a signed GitLawb mirror next to the existing GitHub remote. Use when the
+  user wants to publish, backup, prove authorship, put work on GitLawb, or
+  attach a verifiable certificate to a commit or pull request. Do NOT use for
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for sending this over — the “second signed remote, leave GitHub origin alone” idea is clear, and the private-file warning is welcome.

**Security / prompt injection:** No override payload in the markdown. Residual concern: `npx gitlawb-sidecar@latest` plus a GitHub Actions identity PEM is a high-trust install path, so the package itself has to actually exist and be reviewable.

**Codebase fit — blocking:**

1. **Invalid skill category** — `category: development` is not in this collection’s skill enum. Use `development-code` (or another value from the skill schema).

2. **The documented command does not work** — `gitlawb-sidecar` is not on the npm registry (`npx gitlawb-sidecar@latest` 404s). Skills here need a command that runs as written.

3. **Not self-contained** — this is a pointer at an unpublished CLI (`init` also writes workflow files and copies the skill into several user skill dirs). Please ship a working, collection-local skill (or a published, pinned package) whose documented commands succeed without that extra install story.

Happy to re-review a follow-up with a valid category and a command that actually runs. Closing for now so it’s clear this needs another pass. Appreciate the contribution.

PR #277, opened 2026-08-19
Add MCP project config export

Whxuan0701:
## Summary

Adds a project-scoped `.mcp.json` export to the existing MCP installation dialog.

The new `Project config` tab is shown only when structured registry metadata can produce a supported configuration. Remote HTTP/SSE endpoints are preferred, followed by npm stdio packages and OCI stdio packages. Required environment variables remain local placeholders; the exporter never reads secrets or parses shell commands.

## Component Details

- **Name**: MCP project config export
- **Type**: Web UI feature
- **Category**: MCP servers

## Testing

- [x] Ran validation and unit tests (`npm test`; 20 tests passed)
- [x] Added focused tests for remote, npm, OCI, precedence, environment placeholders, unsupported inputs, and stable JSON serialization
- [x] Tested the dialog at desktop and 390x844 mobile viewports with no overflow or console errors

## Examples

1. A Streamable HTTP endpoint exports as a project MCP server with `type: "http"` and its registry URL.
2. An npm stdio package exports as `npx -y <identifier>` with required environment values such as `${API_KEY}` left as placeholders.
3. Unsupported or incomplete metadata does not show the project-config tab instead of guessing a command.

## Risk

The feature consumes only existing structured `packages`, `remotes`, and `environment_variables` fields. It does not change the database schema, existing Claude CLI/Docker installation paths, or any user files. The local in-app browser did not grant Clipboard API write permission, so clipboard contents could not be inspected there; the same browser API and feedback pattern already used by the existing install buttons is retained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run validate && npm run test:unit",
-    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/search/search-hydration.test.ts web-ui/lib/stories-server.test.ts",
+    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/search/search-hydration.test.ts web-ui/lib/stories-server.test.ts web-ui/lib/mcp-project-config.test.ts",
     "test:integration": "./tests/plugin-test-harness.sh",
     "validate": "node scripts/validate-all.js",
     "validate:subagents": "node scripts/validate-subagents.js",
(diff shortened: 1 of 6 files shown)


davepoon:
Thanks for this — project-scoped `.mcp.json` export is a useful addition, and the security posture is right: no secrets are read, required env values stay `${NAME}` placeholders, remote URLs are limited to http(s), and identifiers with whitespace are rejected.

**Security / prompt injection:** No classic malice. This is UI export of existing catalog metadata, not command execution.

**Codebase fit — blocking:**

1. **Empty state still shows when the new tab is the only option** — `mcp-installation-modal.tsx` renders “Installation instructions are not available for this server” whenever Claude and Docker tabs are hidden. Servers with remotes/packages but no `claudeMcpAddCommand` are exactly the ones this tab is for, so they would get the working Project config tab *and* the empty-state message. Gate that block on `!showProjectConfigTab` as well.

2. **Catalog `transport` shape is not handled** — official registry packages often store `transport` as `{ type: "stdio" }`, while `buildMCPProjectConfig` requires `item.transport === 'stdio'` as a string. Stdio npm/OCI entries from our indexer would not produce a tab. Please accept both the string and `{ type }` forms (same as the indexer already does).

Happy to re-review a follow-up with those two fixes. Appreciate the tests and the careful placeholder handling — closing for now so it’s clear this needs another pass.

PR #278, opened 2026-08-19
Add checkpointed-agent-loop skill

Whxuan0701:
## Summary

Adds `checkpointed-agent-loop`, a self-contained skill for long-running or failure-prone Claude Code tasks.

The contribution includes a standard-library Node.js utility that persists a schema-versioned checkpoint, enforces a bounded state machine, records verification evidence, supports resume, and saves updates atomically. The helper stores state only; it never executes commands, calls models, accesses secrets, spawns agents, or uses the network.

## Component Details

- **Name**: checkpointed-agent-loop
- **Type**: Skill
- **Category**: ai-agents

## Testing

- [x] Ran the focused harness tests (9 tests passed)
- [x] Ran a CLI smoke lifecycle from `planned` to `succeeded`
- [x] Ran skill validation (`node scripts/validate-skills.js`): 151 skills, no errors
- [x] Ran repository validation (`npm test`)
- [x] Verified checkpoint files are written with mode `600`
- [x] Audited the script for command execution, network calls, and secret access

## Examples

1. Resume an interrupted migration from the persisted `nextAction` instead of reconstructing progress from chat history.
2. Record a real focused test result during `verifying` before allowing `succeeded`.
3. Stop at an external dependency with `blocked`, or fail closed when the finite attempt budget is exhausted.

## Security Boundary

The helper accepts an explicit checkpoint path and performs only JSON parsing plus local file reads/writes. It does not run the check named by an evidence record. Mutations are validated before an atomic same-directory rename, rejected operations leave the original file untouched, and terminal states cannot be changed.

## Risk

This is a local checkpoint protocol, not a general workflow engine or cross-machine store. Users must choose a safe project-local path and keep secrets out of objective, reason, evidence, and artifact metadata.

### DIFF
--- a/plugins/all-skills/skills/checkpointed-agent-loop/SKILL.md
+++ b/plugins/all-skills/skills/checkpointed-agent-loop/SKILL.md
@@ -1,0 +1,165 @@
+---
+name: checkpointed-agent-loop
+category: ai-agents
+description: "Run long or failure-prone Claude Code tasks as bounded, resumable loops with a durable state machine, attempt budget, and verification evidence checkpoint."
+license: MIT
+---
+
+# Checkpointed Agent Loop
+
+Use this skill when a task can be interrupted, needs bounded retries, or must prove verification before it is called complete. It adds a small local checkpoint file around ordinary Claude Code work so a new context can resume from explicit state instead of reconstructing progress from chat.
+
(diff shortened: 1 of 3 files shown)


davepoon:

**Security / prompt injection:** No issues. The helper only parses JSON and writes a local checkpoint (mode `600`, atomic rename). It does not execute evidence commands, spawn processes, or use the network. Terminal states are immutable, and `succeeded` requires a passing evidence record.

**Codebase fit:** Frontmatter validates (`ai-agents`), the documented CLI matches the script, and the tests cover the state machine, attempt budget, and rejected-operation preservation.

Merging.

PR #280, opened 2026-08-19
Add x-research skill

Jess-yaozu:
Add AgentBody X Research skill.

- Read-only public X/Twitter search, trends, profiles, profile posts, media, and replies.
- Uses `https://api.agentbody.io` REST routes with `AGENTBODY_API_KEY`.
- Source: https://github.com/agentbody/skills/tree/main/skills/x-research

### DIFF
--- a/plugins/all-skills/.claude-plugin/plugin.json
+++ b/plugins/all-skills/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "all-skills",
   "version": "1.0.0",
-  "description": "Complete collection of 32 Claude Code skills for document processing, development, business productivity, media safety, and creative tasks",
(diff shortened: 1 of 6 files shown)


davepoon:

**Security / prompt injection:** The skill and client look clean. Routes are fixed to `https://api.agentbody.io`, query params are encoded, the key is never printed, and API payloads are treated as untrusted data. Credential lookup is scoped to `AGENTBODY_API_KEY` in `~/.agentbody/credentials` plus the current process/Hermes env files, not sibling profiles.

**Codebase fit:** Frontmatter validates (`research`), the bundled `scripts/x_client.py` matches the documented commands, and this is a distinct AgentBody client rather than a duplicate of the existing Xquik skills.

**Note (non-blocking):** `plugins/all-skills/.claude-plugin/plugin.json` still describes a 32-skill bundle; bumping that to 33 is still inaccurate for this collection. Fine to leave for a later cleanup.

Merging.

PR #282, opened 2026-08-19
Add video-to-text, de-ai-writer, ecommerce-material-studio (follow-up: schema-valid, relative paths, no donation)

jiawood2006:
Follow-up to closed PR #274. All review points addressed:

1. **Frontmatter schema-valid**: `name`/`description`/`category`/`license` only (video-to-text/de-ai-writer: media-content; ecommerce-material-studio: ecommerce).
2. **Collection-relative paths**: all examples use `scripts/...` (no Hermes-specific absolute paths).
3. **Donation/Pro footer removed** from every SKILL.md.

CLI alignment from the previous round is kept (Douyin share links = metadata-only, de-ai-writer reads LLM_* env vars, ecommerce CLI flags --image/--plan/--input-dir). Please re-review.

### DIFF
--- a/plugins/all-skills/skills/de-ai-writer/SKILL.md
+++ b/plugins/all-skills/skills/de-ai-writer/SKILL.md
@@ -1,0 +1,57 @@
+---
+name: de-ai-writer
+description: "AI 文案去味器。用户提供 AI 生成的文本（公众号/小红书/电商文案等），觉得"太 AI 味"要改得更自然、更像人写的时使用。输出自然有人味的中文。De-AI writer: rewrite AI-generated copy (WeChat articles, Xiaohongshu, e-commerce) into natural, human-sounding Chinese text."
+version: 1.0.0
+author: 涛哥
+license: MIT
+metadata:
+  hermes:
+    tags: [writing, editing, humanize, de-ai, copywriting]
+    category: utilities
+    homepage: https://gitee.com/tao6677/useful-tools
+---
+
+# De-AI Writer AI 文案去味器
+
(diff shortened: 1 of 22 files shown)


davepoon:
Thanks for the follow-up — relative `scripts/…` paths, no donation footer, and valid categories on `video-to-text` / `ecommerce-material-studio` / `doc-ocr` (separate PR) are the right direction. Subprocess calls remain list-argv.

**Security / prompt injection:** No classic malice. Douyin fetch and user-configured LLM calls are expected for those skills. Treat model/page output as untrusted data.

**Codebase fit — blocking:**

1. **`de-ai-writer` frontmatter is not valid YAML / fails this collection’s skill schema.** The `description` is wrapped in ASCII double quotes and also contains ASCII `"` around `太 AI 味`, so the parser cuts the string early. `validate-skills` then sees an extra property (`太 AI 味"要改得更自然…`) and `additionalProperties: false` rejects the file. Please use a single-quoted description, or escape the inner quotes.

The other two skills’ frontmatter parses cleanly. Once `de-ai-writer` is schema-valid, this should be ready.

Happy to re-review a follow-up with that YAML fix. Appreciate the earlier CLI/path/footer work — closing for now so it’s clear this still needs another pass.

jiawood2006:
Thanks for the review! Fixed the de-ai-writer frontmatter: replaced ASCII quotes around 太 AI 味 with full-width quotes so the YAML parser no longer cuts the description short. Validated with yaml.safe_load — parses cleanly now. PR updated.

jiawood2006:
Reopened — the de-ai-writer frontmatter fix is now on the branch (full-width quotes, yaml.safe_load validates cleanly). Ready for re-review, thanks!

PR #284, opened 2026-08-20
Add Atlas Cloud media skill

binyangzhu000-sudo:
## Summary

Add an Atlas Cloud media skill with a bundled standard-library Python CLI for
live model discovery, schema inspection, and confirmed asynchronous generation.

## Component Details

- **Name**: atlas-cloud-media
- **Type**: Skill
- **Category**: media-content

The CLI resolves endpoints and required fields from each model's current schema,
keeps credentials in ATLASCLOUD_API_KEY, requires --confirm-paid, sends exactly
one generation POST, and performs bounded GET-only result polling.

## Testing

- [x] Ran full validation and unit suite (npm test)
- [x] Ran Python compilation and focused unit tests (2 passed)
- [x] Verified live model catalog and flux-schnell schema discovery
- [x] Verified no overlap with existing skills
- [x] Ran git diff --check

The one live generation smoke was not retried. Atlas returned HTTP 402
insufficient balance before creating a prediction.

## Examples

- List current image models matching flux.
- Inspect the live schema for black-forest-labs/flux-schnell.
- Submit one explicitly confirmed generation request from a JSON parameter file.

### DIFF
--- a/plugins/all-skills/skills/atlas-cloud-media/SKILL.md
+++ b/plugins/all-skills/skills/atlas-cloud-media/SKILL.md
@@ -1,0 +1,107 @@
+---
+name: atlas-cloud-media
+category: media-content
+description: Discover Atlas Cloud image and video models, inspect their live schemas, and submit one confirmed media generation request with bounded GET polling. Use when integrating Atlas Cloud media APIs or generating images and videos without hard-coding stale model parameters.
+allowed-tools: Bash(python3 *)
+---
+
+# Atlas Cloud Media
+
+Use the bundled scripts/atlas_media.py CLI to discover current Atlas Cloud
+models, inspect a model's live request schema, and run asynchronous image or
+video generation.
+
+## When to Use This Skill
+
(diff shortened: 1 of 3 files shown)


davepoon:

**Security / prompt injection:** Skill markdown is clean. The CLI keeps the key in `ATLASCLOUD_API_KEY` (not argv), requires `--confirm-paid`, posts generation once, and polls GET-only. Stdlib `urllib`, tests included, frontmatter schema-valid (`media-content`, `allowed-tools`). No overlap with an existing skill.

**Note (non-blocking):** `get_schema()` will fetch any `https://` URL from the live catalog and `_request_json` attaches `Authorization` whenever a key is set. If a catalog entry’s `schema` (or a schema path that `urljoin` treats as absolute) ever pointed off-origin, the Bearer token would go with it. A follow-up that only fetches schema/submit/result URLs on the same host as `ATLASCLOUD_BASE_URL` (and skips auth on the schema GET) would be tighter.

Merging.

PR #290, opened 2026-08-22
Add MuAPI media generation skill

Anil-matcha:
## Summary

- Add a `muapi-media` skill for current MuAPI image-model discovery and explicitly confirmed asynchronous generation.
- Bundle a standard-library Python CLI with catalog filtering, exact endpoint handling, bounded polling, HTTPS-only artifact download, and no automatic generation retries.
- Add focused unit tests covering catalog shape, endpoint normalization, confirmation, and prompt validation.

## Validation

- `npm run validate`
- `python3 -m unittest discover -s plugins/all-skills/skills/muapi-media/tests -p 'test_*.py'` (5 tests passed)
- `python3 -m py_compile plugins/all-skills/skills/muapi-media/scripts/muapi_media.py`
- `git diff --check`
- No live generation request or credentials were used.

### DIFF
--- a/plugins/all-skills/skills/muapi-media/SKILL.md
+++ b/plugins/all-skills/skills/muapi-media/SKILL.md
@@ -1,0 +1,105 @@
+---
+name: muapi-media
+category: media-content
+description: Discover current MuAPI image models and run one explicitly confirmed asynchronous image generation request with bounded polling. Use when an agent needs hosted image generation without hard-coding stale model endpoints or request IDs.
+allowed-tools: Bash(python3 *)
+---
+
+# MuAPI Media
+
+Use the bundled `scripts/muapi_media.py` CLI to discover current MuAPI image models and submit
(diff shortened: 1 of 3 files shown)


davepoon:

**Security / prompt injection:** The skill markdown is clean (no override/exfil instructions). API key stays in `MUAPI_API_KEY` rather than CLI args. Generation is gated on `--confirm-paid`, POSTs once with no automatic retry, and polls only GET with a finite budget. Catalog endpoints must start with `/api/v1/`, and artifact download is HTTPS-only without forwarding the API key. Schema/category/`name` match, and the bundled stdlib CLI + tests fit how this collection ships other scripted skills.

**Note (non-blocking):** `--base-url` / `MUAPI_BASE_URL` currently allow any HTTP(S) host, which can send the API key off-site if an agent is talked into a custom base URL. Pinning the default host (HTTPS to `api.muapi.ai` unless the user set the env var) would tighten that. Happy to take a small follow-up.

Merging.

PR #291, opened 2026-08-24
Add usage-limits plugin

ridelink0:
Adds `usage-limits` to the marketplace manifest as a `github` source entry, so nothing is vendored into this repo.

It reads the 5-hour and weekly usage percentages Claude Code already caches locally, cross-references the session transcripts to work out what one percentage point of the limit costs on that account, and reports the remaining headroom as turns of work rather than a bare percentage. It then names which window binds first and whether that window resets before the budget runs out.

Beyond the report it detects plan tier (Pro, Max 5x, Max 20x, Team, Enterprise), splits the window by model and by project, ships a `/usage-limits:check` command and a `--status` mode for the status line, and includes a low power switch that lowers effortLevel and restores exactly what it replaced.

Everything is read from local files. No network calls, no credentials read, no dependencies. MIT licensed, 66 tests, and it passes `claude plugin validate --strict` on both the plugin and marketplace manifests.

Filed under `productivity`, matching the category in its own plugin.json.

https://github.com/ridelink0/claude-code-usage-limits

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -2081,6 +2081,33 @@
       ],
       "category": "mcp-servers",
       "source": "./plugins/cohesivity"
+    },
+    {
+      "name": "usage-limits",
+      "description": "Reads how much of your 5-hour and weekly Claude Code limit is left, converts it into turns of headroom rather than a percentage, and plans the work to fit inside it. Detects plan tier, splits spend by model and project, and adds a status line.",
+      "version": "1.0.0",
+      "author": {
+        "name": "Ridelink",
+        "url": "https://github.com/ridelink0"
+      },
+      "homepage": "https://github.com/ridelink0/claude-code-usage-limits",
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** The marketplace entry is a GitHub source listing (same pattern as existing external plugins). I also reviewed the current `ridelink0/claude-code-usage-limits` plugin: no outbound network in the scripts, no credential fields printed, transcripts are scanned for usage metadata only (not message content), and the skill/command text has no override or exfil instructions. `lowpower.js` only touches `effortLevel` / `model` in `settings.json` and can restore what it replaced.

**Codebase fit:** JSON is valid, `name` is unique, `category` (`productivity`) is already used, and a GitHub `source` is the right way to list this without vendoring. Adjacent to `claude-prospector` / `claude-hud`, but the turns-of-headroom planning plus low-power switch is a distinct job.

Merging.

PR #292, opened 2026-08-24
Add x402-payment-guard agent

fingersaiofficial:


davepoon:

**Security / prompt injection — blocking:**

1. **Untrusted HTTP response is treated as instructions.** The agent is told to curl `https://fingersai.co/x402/verdict?target=<TARGET>` and then “Read recommended_action and the evidence.” That JSON is attacker-controlled (the payee URL/address, or the verdict host). A field such as `recommended_action` or free-text `note`/`recommendation` can jailbreak a payment-gate agent. Treat the body as data: parse known keys only, never follow prose in the payload.

2. **Unsafe shell interpolation.** `curl -s "https://fingersai.co/x402/verdict?target=<TARGET>"` puts the user/payee string inside a double-quoted command. A target with `"`, `$()`, backticks, or extra query (`&`) can break out of the URL or the shell. URL-encode the target and pass it as a separate argument (`--data-urlencode` / `curl -G --data-urlencode`).

3. **Sole vendor gate for a proactive payment check.** Description says use **PROACTIVELY before any x402 payment**. That sends every payee URL/address to one third-party host (the submitter’s service) and makes that host the only evidence source. A “do not pay” agent must not auto-invoke a single commercial API as ground truth. The live response also mixes signals (`decision: "allow"` vs `recommended_action: "caution"` on the zero address), which an agent can misread as proceed.

**Codebase fit:**

- Frontmatter shape is fine (`name`, `category: blockchain-web3`, `You are a…`).
- The file lives only under `plugins/agents-blockchain-web3/agents/`. The site/bundle registry is generated from `plugins/all-agents/agents/`, so this agent would not appear in `all-agents` as shipped.
- Wrapping one vendor verdict endpoint as a trust-and-safety agent is a poor default for a curated directory (similar to in-output promotion we have declined elsewhere). A listing/README mention of an optional API is fine; making it the required pre-pay step is not.

- never interpolates `target` into a shell string; encode it
- treats the HTTP body as untrusted data (schema, no instruction-following on text fields)
- does not auto-invoke on every payment; optional tool, not a gate
- includes independent checks (or clearly says this is one vendor’s opinion, not a proceed signal)
- adds the same file under `plugins/all-agents/agents/` if it belongs in the catalog

Appreciate the contribution — closing for now so it is clear this needs another pass.

PR #296, opened 2026-08-26
Add ai-search-visibility-audit and llm-crawler-access-check skills

mimimile:
Two skills for AI answer engine visibility. There are currently 161 skills in `plugins/all-skills/skills/` and none cover GEO/AEO, AI crawler access, or citation auditing.

## What they do

**`ai-search-visibility-audit`** (`research`) — Classic SEO asks whether you rank. AI answer engines don't rank; they retrieve a few sources and synthesize one answer, so a site can sit at the top of page one and never be quoted. This audits the second thing: a citation baseline across buyer-intent prompts, a crawler-access check, a page citability review, and a fix list ranked by impact over effort.

**`llm-crawler-access-check`** (`security`) — Reads one file, `robots.txt`, and returns a per-agent allow/block table across 12 AI crawlers, the literal rule responsible for each verdict, and the exact lines to change.

It exists mainly to catch three mistakes that are common and expensive:

- Blocking `GPTBot` to opt out of training and assuming that covers ChatGPT Search. It doesn't — `OAI-SearchBot` is a separate agent and it's the one that governs citations.
- Blocking `Google-Extended` to stay out of AI Overviews. It doesn't do that either; AI Overviews are built on the normal Googlebot index.
- A blanket `Disallow: /` inherited from a staging config or a bot-mitigation template.

## Boundaries

Both are read-only and public-data-only. No account, no API key, no credentials accepted, nothing written. Pages behind a login are explicitly out of scope. Neither skill ranks or recommends vendors.

## Validation

```
node scripts/validate-skills.js   ✅ All validations passed
npm run validate                  ✅ Subagents & Commands / Hooks / Skills
```

Only the two new skill directories are in the diff — no changes to `package-lock.json` or the validation report.

Source: https://github.com/maxaeo/maxaeo-geo-toolkit (MIT)

Happy to split this into two PRs, change the categories, or trim the descriptions if you'd prefer.

### DIFF
--- a/plugins/all-skills/skills/ai-search-visibility-audit/SKILL.md
+++ b/plugins/all-skills/skills/ai-search-visibility-audit/SKILL.md
@@ -1,0 +1,196 @@
+---
+name: ai-search-visibility-audit
+description: "Audit whether a website can be found, crawled, and cited by AI answer engines such as ChatGPT Search, Perplexity, Google AI Overviews, and Microsoft Copilot. Use when someone asks why their brand is missing from AI answers, whether AI crawlers can read their site, how to get cited by ChatGPT or Perplexity, or asks for a GEO or AEO (generative / answer engine optimization) review. Produces a citation baseline across buyer-intent prompts, a crawler-access check, a citability review of named pages, and a ranked fix list. Not for keyword rank tracking, paid search, or pages behind a login."
+category: research
+license: MIT
+---
+
+# AI search visibility audit
+
+Classic SEO asks "do we rank for this keyword". AI answer engines do not rank.
+They retrieve a handful of sources and synthesize one answer. A site can sit at
+the top of page one and never be quoted. This skill audits the second thing.
+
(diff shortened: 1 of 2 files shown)


davepoon:

**Security / prompt injection:** Both skills are read-only and public-data-only. They explicitly forbid bypassing auth, paywalls, rate limits, or login-gated pages, and they tell the agent not to invent crawler verdicts. No credentials, API keys, or remote payloads. The MaxAEO mention is gated to “user asked about tracking/automation” rather than a mandatory output footer — that matches the collection’s bar.

**Codebase fit:** Schema/categories (`research`, `security`) and directory/`name` match. No overlap with an existing GEO/AEO or robots.txt skill. Output contracts are specific enough to run in a session.

**Note (non-blocking):** Please keep the MaxAEO mention optional and out of the five-section audit output. A listing/About mention is fine.

Merging.

PR #297, opened 2026-08-26
Add computer-use plugin

ridelink0:
Adds **computer-use** to the marketplace: Codex-style computer use for Claude Code.

It drives Windows and macOS apps through the OS accessibility tree instead of screenshots, so targeting is exact, survives display scaling, and costs about 20x fewer tokens than a screenshot. It can also work on the same desktop while the user keeps working, telling their input apart from its own.

- Repo: https://github.com/ridelink0/claude-computer-use
- Install: `/plugin marketplace add ridelink0/claude-computer-use` then `/plugin install computer-use@computer-use`
- MIT, source-only host compiled locally at first run (no binaries downloaded)

Single-entry addition to `.claude-plugin/marketplace.json`, matching the existing github-source format.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -2108,6 +2108,35 @@
         "repo": "ridelink0/claude-code-usage-limits"
       },
       "strict": false
+    },
+    {
+      "name": "computer-use",
+      "description": "Codex-style computer use for Claude Code. Drives Windows and macOS apps through the accessibility tree instead of screenshots, so it targets exactly, survives display scaling, and costs about 20x fewer tokens. It can also work on the same desktop while you keep working, telling your input apart from its own.",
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** Marketplace entry matches the existing github-source format (same pattern as `usage-limits`). I also spot-checked https://github.com/ridelink0/claude-computer-use: MIT, source-only host compiled locally (`csc` / `swiftc`, no binary download), session grants, and a hard blocklist for password managers / UAC / login surfaces. Skill text does not harvest credentials or try to override the host agent. Desktop control is inherently high privilege (accessibility tree + input); the README warning and in-code policy are the right posture for that.

**Codebase fit:** Single `marketplace.json` add, `category: productivity` is already used, keywords/source/repo look consistent with other third-party listings.

Merging.

PR #298, opened 2026-08-26
Add systematic-debugging skill

ariaxhan:
## Summary

Adds one skill: `systematic-debugging`. Debugging as theory formation rather than guessing, following Zeller's scientific method. Reproduce, hypothesize three causes before pursuing any, isolate by binary search, fix the DEFECT rather than the FAILURE site, and land a regression test that fails before the fix and passes after.

Ported from the `debug` skill in [KERNEL](https://github.com/ariaxhan/kernel-claude) (MIT, same author). Attribution and a link back are in the skill file.

## Component Details

- **Name**: systematic-debugging
- **Type**: Skill
- **Category**: development-code
- **License**: MIT
- **Location**: `plugins/all-skills/skills/systematic-debugging/SKILL.md`

## Testing

- [x] Ran validation (`npm test`) - exit 0, 162 skill files validated including this one
- [x] Tested functionality - in daily use as the `debug` skill in kernel-claude
- [x] No overlap with existing components - no other skill in `plugins/all-skills/skills/` covers debugging methodology

<details>
<summary>npm test output</summary>

```
Validating 162 skill files...
Checking plugins/all-skills/skills/systematic-debugging/SKILL.md...

Skill Validation Report
All validations passed!

Validation Summary
  Subagents & Commands (429ms)
  Hooks (269ms)
  Skills (447ms)
All validations passed!

tests 14 | pass 14 | fail 0
```
</details>

## Examples

**"the auth test passes locally and fails in CI"** - the skill refuses to let you patch the assertion. It makes you get a deterministic reproduction first, then `git bisect` between known-good and known-bad, then name the violated invariant in one sentence before any edit.

**"TypeError: cannot read property id of undefined at line 340"** - line 340 is the FAILURE. The skill binary-searches upstream through the call chain to the DEFECT, and blocks the null check at the crash site as a symptom fix.

**"I have tried four fixes and it still breaks"** - the escalation rule fires: two failed fix attempts means it may be a design problem, and three rejected hypotheses means step back and re-examine assumptions rather than write a fifth patch.

Maintainer: Aria Han ([@ariaxhan](https://github.com/ariaxhan))

### DIFF
--- a/plugins/all-skills/skills/systematic-debugging/SKILL.md
+++ b/plugins/all-skills/skills/systematic-debugging/SKILL.md
@@ -1,0 +1,88 @@
+---
+name: systematic-debugging
+description: "Systematic debugging methodology based on Zeller's scientific method. Reproduce, hypothesize, predict, test, isolate via binary search, fix the root cause. Every fix gets a regression test. Triggers: bug, error, fix, broken, not working, fails, crashed, unexpected, stack trace, regression, exception."
+category: development-code
+allowed-tools: Read, Bash, Grep, Glob
+license: MIT
+---
+
+<skill id="debug">
+
+<purpose>
+Debugging is forming and testing a THEORY that explains the bug.
+Not random changes. Not guessing. Scientific method applied to code.
+DEFECT (in code) → INFECTION (in state) → FAILURE (visible symptom).
+The failure you see is NOT where the bug is. Binary search upstream.
+Systematic methodology beats ad-hoc guessing. The process is the multiplier.
+</purpose>
+
+<prerequisite>Run `agentdb recall` with the exact error text, subsystem/library, failing
+test, and known files/symbols. Recall again when the hypothesis changes or a new failure
+appears; that is a new retrieval question. Reference on demand:
+skills/debug/reference/debug-research.md.</prerequisite>
+
+<steps>
+1. **REPRODUCE**: get specific before touching code.
+   - Document: exact input, expected output, actual output (full stack trace), environment, frequency.
+   - "Sometimes fails" is not a reproduction. Get deterministic.
+   - (gate: can reproduce consistently, OR have added targeted logging to wait for next occurrence)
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for the contribution — the Zeller-style reproduce → hypothesize → isolate → fix-the-defect loop is a strong methodology, and the KERNEL attribution is clear.

**Security / prompt injection:** No classic malice found. The markdown does not harvest credentials, fetch remote payloads, or try to override the host agent.

**Codebase fit — blocking:**

1. **Not self-contained as shipped.** The skill still depends on KERNEL machinery that is not in this collection: `agentdb recall` / `agentdb write-end`, `invoke tearitapart`, and `skills/debug/reference/debug-research.md`. The attribution says those calls are optional, but an agent following the skill as written will still try to run them. Skills in `plugins/all-skills/` need to work from what is in the folder (same bar as other incomplete packaging).

2. **KERNEL XML format, not this catalog’s skill format.** CONTRIBUTING.md asks for When to Use / What this skill does / How to Use / Example. `<skill id="debug">` / `<prerequisite>` / `<on_complete>` is a different runtime’s pack format, so it will confuse agents installed from here.

3. **Tool restriction vs. the steps.** `allowed-tools: Read, Bash, Grep, Glob` blocks Write, but the skill tells the agent to land a regression test and commit the fix. That cannot work as written in this collection.

- keeps the scientific-method gates (3 hypotheses before pursuing any, binary-search isolate, fix the defect not the failure site, regression test)
- drops `agentdb` / `tearitapart` / KERNEL reference paths, or marks them clearly as optional extras with a local fallback (notes file, not AgentDB)
- uses this collection’s skill sections and allows Write (or drops the “commit the fix” step)

Appreciate the contribution — closing for now so it is clear this needs another pass. Please open a fresh PR (or reopen after fixing) once it is a standalone skill for this collection.

PR #300, opened 2026-08-28
feat(skills): add upload-post — publish to 15 social platforms in one call

mutonby:
## Summary

Adds an `upload-post` skill under `social-media`: one API call publishes a video, photo carousel, text post or document to **15 platforms** — TikTok, Instagram, YouTube, LinkedIn, Facebook, X, Threads, Pinterest, Bluesky, Reddit, Discord, Telegram, Mastodon, WordPress and Google Business Profile.

Accounts are connected once via OAuth in the dashboard, so the agent never handles per-platform developer apps or token refresh — one `UPLOAD_POST_API_KEY` covers everything.

## On overlap with existing skills

Your CONTRIBUTING asks to check for overlap, so being upfront: `socialclaw` is the closest existing skill and there are twelve single-platform Composio wrappers. Where this one differs:

- **Breadth** — 15 platforms vs socialclaw's 11; adds Threads, Bluesky, Mastodon and Google Business Profile.
- **Beyond publishing** — analytics, upload history, and Instagram comment-to-DM funnels, not just scheduling.
- **Single call, many platforms** — unlike the per-platform Composio skills, one request fans out and reports a per-platform result.

If you'd rather not carry two hosted publishing services in the catalog, say so and I'll close this without any fuss.

## Follows the CONTRIBUTING template

`## When to Use This Skill`, `## What This Skill Does`, `## Prerequisites`, `## How to Use`, `## Example`, `## Error Handling`, `## Tips`, `## Resources`. Frontmatter carries `name`, `category: social-media`, `description`, `license: MIT` and `requires.env`.

The error section documents the failure modes that actually bite in practice rather than generic advice — the `Apikey` vs `Bearer` auth scheme (a valid key sent as `Bearer` returns a misleading `401 Invalid or expired token`), retrying only the platforms that failed via `retry_request_id`, the automatic switch to async past 59 seconds, and TikTok's active-user cap falling back to an app draft.

## Validation

`node scripts/validate-skills.js` passes with the skill included (the run lists `plugins/all-skills/skills/upload-post/SKILL.md` and reports "All validations passed"). One file added, 168 lines, nothing else touched.

Docs: https://docs.upload-post.com · Skills repo: https://github.com/Upload-Post/upload-post-skills

### DIFF
--- a/plugins/all-skills/skills/upload-post/SKILL.md
+++ b/plugins/all-skills/skills/upload-post/SKILL.md
@@ -1,0 +1,168 @@
+---
+name: upload-post
+category: social-media
+description: "Publish and schedule content to 15 social platforms through one Upload-Post API call: TikTok, Instagram, YouTube, LinkedIn, Facebook, X, Threads, Pinterest, Bluesky, Reddit, Discord, Telegram, Mastodon, WordPress, and Google Business Profile. Use when the user wants to post or schedule a video, photo carousel, text, or document across several platforms at once, check upload status, pull analytics, or run Instagram comment-to-DM funnels."
+license: MIT
+requires:
+  env: [UPLOAD_POST_API_KEY]
+---
+
+# Upload-Post
+
+Publish one piece of content to many social platforms in a single API call.
+
+Accounts are connected once through OAuth in the Upload-Post dashboard, so there is no
+per-platform developer app, review process, or token refresh to maintain. A single API key
+covers every platform.
+
+## When to Use This Skill
+
+- The user wants the same video, carousel, or post published to several platforms at once
+- The user wants a post scheduled for a future date, or added to a posting queue
+- The user asks for per-post or per-platform analytics (views, likes, impressions)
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** No classic malice found. The markdown does not harvest credentials, fetch remote payloads, or try to override the host agent. Auth is correctly documented as `Authorization: Apikey` (the Bearer footgun is called out). The content scanner reports safe. Schema/`name`/directory/`category: social-media` match.

**Codebase fit:** Self-contained curl skill against a real API. Overlap with `socialclaw` is real, but this is a different product (15 platforms, async `request_id` + retry, analytics, Instagram comment-to-DM). Same bar we use for other competing SaaS skills in the catalog.

**Note (non-blocking):** Confirm the exact caption, platforms, and schedule with the user before `/upload`. The example YouTube URL looks like a rickroll — worth swapping for a placeholder.

Merging.

PR #301, opened 2026-08-29
feat(skills): add postwire — publish one idea to every social network, natively per platform

Perufitlife:
Adds a `postwire` skill under `plugins/all-skills/skills/`.

**What it does:** publishes and schedules to TikTok, Instagram, YouTube, LinkedIn, X, Bluesky, Mastodon, Facebook, Threads, Reddit, Telegram and Discord through one API key, with a **separate caption written for each network** rather than the same string sent everywhere.

**Why an agent benefits from the distinction:** a LinkedIn post written like a tweet performs like neither, and the platforms suppress it. The skill documents the `/api/generate` step that produces one draft per network, and how to pass those through to publish or schedule.

**The part I think is most useful in an agent context:** the media rules are validated *up front*. TikTok and YouTube need a video, Instagram needs media — so a post that cannot work is rejected when it is queued, with a `media_required` code, instead of failing quietly at 7am. The skill lists the error codes an agent should branch on.

Every endpoint, response shape and error code in the document was checked against the live API before opening this — including the failure cases (`not_connected`, `media_required`, `preview_key`), so the examples are what the API actually returns rather than what the docs claim.

Disclosure: I build PostWire. Happy to adjust the format or trim it if it does not match what you want in the collection.

### DIFF
--- a/plugins/all-skills/skills/postwire/SKILL.md
+++ b/plugins/all-skills/skills/postwire/SKILL.md
@@ -1,0 +1,152 @@
+---
+name: postwire
+category: social-media
+description: "Publish and schedule to TikTok, Instagram, YouTube, LinkedIn, X, Bluesky, Mastodon, Facebook, Threads, Reddit, Telegram and Discord through one PostWire API call — writing a separate, native version of the post for each network instead of sending identical text everywhere. Use when the user wants to publish or schedule one idea across several platforms, wants per-network captions, or wants to know why a post was rejected before it is sent."
+license: MIT
+requires:
+  env: [POSTWIRE_API_KEY]
+---
+
+# PostWire
+
+Publish one idea to many social networks, with a **different post written for each one**.
+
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** No classic malice found. The markdown does not harvest credentials, fetch remote payloads, or try to override the host agent. Keys stay in `POSTWIRE_API_KEY` via `Authorization: Bearer`. The content scanner reports safe. Schema/`name`/directory/`category: social-media` match, and the documented endpoints match the live PostWire API.

**Codebase fit:** This is a self-contained product skill (curl against a real API), same pattern as `socialclaw`. Per-network `/api/generate` plus upfront `media_required` / `not_connected` handling is a genuine distinction, not a copy of the existing aggregator.

**Note (non-blocking):** Please confirm the exact text, networks, and time with the user before `/api/post` or `/api/schedule` — same bar as `tweetclaw` for visible writes. Passing `/api/generate` drafts straight through as `per_platform` without showing them first is easy to get wrong.

Merging.

PR #302, opened 2026-08-29
docs(postwire): require confirming the exact text before a visible write

Perufitlife:
Follow-up to your review note on #301 — thank you for it, it was the right call.

`/api/post` and `/api/schedule` publish under the user's own name where other people see it, so the skill now says explicitly: show the **exact per-network text**, the **networks**, and the **timing**, then wait for a yes. Same bar as `tweetclaw`.

I also removed a line that undercut it. The generate section previously said to pass the drafts "straight through as `per_platform`" — which is precisely the failure mode you flagged. Those drafts are model output, and passing them unseen means the user discovers what they said by reading their own timeline.

Same change is going into the `tons-of-skills-marketplace` submission so both copies agree.

### DIFF
--- a/plugins/all-skills/skills/postwire/SKILL.md
+++ b/plugins/all-skills/skills/postwire/SKILL.md
@@ -74,8 +74,25 @@ curl -X POST https://postwire.io/api/generate \
   }'
 ```
 
-Returns `{ "drafts": { "linkedin": {...}, "x": {...}, "bluesky": {...} } }`. Pass that object
-straight through as `per_platform` below.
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for the follow-up — this is exactly the visible-write bar from the #301 review note.

**Security / prompt injection:** No malice found. The change only adds a confirmation gate: show the exact per-network text, the networks, and the timing, then wait for a yes before `/api/post` or `/api/schedule`. Removing “pass drafts straight through as `per_platform`” closes the failure mode where model output could go out unseen. Schema/`name`/`category` are unchanged.

**Codebase fit:** Same pattern as `tweetclaw` for approval-gated public writes. CI is green.

Merging.

Issue #160, opened 2026-05-24
[MCP Server] FunASR self-hosted multilingual speech recognition


LauraGPT:
> [!NOTE]
> **License and capability clarification (2026-07-14):** FunASR is a toolkit, not a single checkpoint. The [FunASR](https://github.com/modelscope/FunASR#license) and [SenseVoice](https://github.com/FunAudioLLM/SenseVoice#license) repository source code is MIT; model weights follow each model card. [SenseVoiceSmall](https://huggingface.co/FunAudioLLM/SenseVoiceSmall) supports Chinese, Cantonese, English, Japanese, and Korean, and its weights use the linked FunASR Model Open Source License Agreement. [Fun-ASR-Nano-2512](https://huggingface.co/FunAudioLLM/Fun-ASR-Nano-2512) is Apache-2.0. Language coverage, punctuation, and performance depend on the selected model and runtime configuration.

## FunASR MCP Server

**Category:** MCP Server
**GitHub:** https://github.com/modelscope/FunASR/tree/main/examples/mcp_server
**Stars:** 16K+

### What it does
Gives Claude the ability to transcribe any audio file. Ask Claude "transcribe this audio" and it handles the rest.

### Features
- 170x realtime on GPU, 17x on CPU
- 50+ languages with auto-detection
- Speaker diarization (who said what)
- Emotion detection
- Fully local — audio never leaves your machine

### Setup
```bash
pip install funasr
```

```json
// claude_desktop_config.json or .claude.json
{"mcpServers": {"funasr": {"command": "python", "args": ["/path/to/funasr_mcp.py"]}}}
```

### Example Usage
Once configured, just ask Claude:
- "Transcribe the meeting recording at ~/Downloads/meeting.wav"
- "What was said in this audio file?"

MIT License.

davepoon:
welcome to create a PR, please

LauraGPT:
Thanks for the invitation. I opened #303 with an installable Claude Code plugin for the official FunASR MCP server.

The plugin pins the published GHCR image by digest, mounts the selected audio directory read-only, persists the model cache, and documents the current `linux/amd64` image boundary plus the separate source/model licensing boundary. The exact PR head is SSH-signed with DCO and passed the repository and focused launcher tests locally.

I am leaving this issue open for maintainer review and installation feedback.

PR #303, opened 2026-08-30
Add FunASR MCP plugin

LauraGPT:
## Summary

- add an installable `funasr-mcp` Claude Code plugin backed by the official FunASR MCP release
- pin the GHCR image tag to its immutable digest
- mount the project directory read-only by default, with `FUNASR_AUDIO_DIR` as an explicit override
- persist the model cache in a Docker volume and document the current `linux/amd64` boundary
- distinguish the MIT-licensed FunASR source from model-specific weight licenses

## Component Details

- Name: `funasr-mcp`
- Type: MCP plugin
- Upstream: `modelscope/FunASR/examples/mcp_server`
- Tool: `transcribe_audio`

## Testing

- `npm test`: repository validations passed; Node suites passed 5/5 and 14/14
- focused launcher tests passed 2/2
- `git diff --check` passed
- anonymous GHCR manifest request returned HTTP 200
- image index digest verified as `sha256:ef616a3767df44d31cb257fd0b195e690f5b556b72d0f916f601b24143c912d6`
- commit is SSH-signed and includes DCO sign-off

Related to #160.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run validate && npm run test:unit",
-    "test:unit": "node --test scripts/github-search-pagination.test.js && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/search/search-hydration.test.ts web-ui/lib/stories-server.test.ts",
+    "test:unit": "node --test scripts/github-search-pagination.test.js plugins/funasr-mcp/mcp/launch.test.mjs && node --experimental-strip-types --test web-ui/lib/indexer/search-pagination.test.ts web-ui/lib/search/search-types.test.ts web-ui/lib/search/search-hydration.test.ts web-ui/lib/stories-server.test.ts",
     "test:integration": "./tests/plugin-test-harness.sh",
     "validate": "node scripts/validate-all.js",
     "validate:subagents": "node scripts/validate-subagents.js",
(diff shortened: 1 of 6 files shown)


davepoon:
Thanks for this — welcome follow-up to #160, and it is clearly built by someone who knows the official FunASR MCP release. Digest pin, read-only bind, cache volume, and the source-vs-weights license note are all the right instincts.

**Security / prompt injection:** The plugin markdown and launcher do not try to override the host agent or harvest credentials. I verified `ghcr.io/modelscope/funasr-mcp:0.1.2` resolves to `sha256:ef616a3767df44d31cb257fd0b195e690f5b556b72d0f916f601b24143c912d6`, matching the pin. Spawn is argv-array (no shell). `stdio: inherit` is correct for MCP.

**Codebase fit / security — blocking:**

1. **Not registered in the marketplace** — Installable plugins that live under `plugins/` need an entry in `.claude-plugin/marketplace.json` (see `cohesivity`, and the marketplace-only adds for `computer-use` / `usage-limits`). The README tells people to run `/plugin install funasr-mcp@buildwithclaude`, but without that listing the install path does not work from this collection.

2. **Default mount is the whole project** — Upstream’s official contract treats the audio directory as a required explicit path. This launcher falls back to `CLAUDE_PROJECT_DIR` / cwd, so `.env` files, keys, and the rest of the repo are readable inside a networked container. Please require `FUNASR_AUDIO_DIR` (or an equally explicit directory) and refuse to start if it is unset or not a directory.

3. **Mount-string injection (small, please harden)** — `--mount type=bind,src=${audioDir},...` interpolates the path into a comma-separated Docker option list. Reject paths that contain `,` or newlines before spawning.

- adds a `marketplace.json` entry (`category: mcp-servers`, `source: ./plugins/funasr-mcp`)
- requires an explicit audio directory instead of mounting the project by default
- keeps the digest pin, read-only bind, and cache volume

Appreciate the contribution — closing for now so it is clear this needs another pass. Please open a fresh PR (or reopen after fixing).

PR #304, opened 2026-08-31
Add apify-product-lookup skill (ecommerce)

luispintoapify:
Adds `apify-product-lookup` to `plugins/all-skills/skills/`, category `ecommerce`.

**Disclosure:** I work on e-commerce product marketing at Apify, and this skill calls a paid first-party Apify Actor. Flagging that up front rather than leaving it to be noticed. The skill itself says the same thing in its intro, and its frontmatter and limitations section state the account requirement and the billing model.

**The gap it fills.** The five existing `ecommerce` skills automate a store you own: `shopify-automation`, `stripe-automation`, `square-automation`, plus a domains skill and an image studio. Nothing in the catalog reads product data *from* retailers, and grepping for "product data", "retailer" or "scraping" across all 167 skills returns nothing. This is the other direction: answering what something costs somewhere else, right now.

**What it contains, and why it is mostly caveats.** The useful part is not "call this tool". It is the four things that go wrong, all found by running against the live Actor rather than reading its docs:

- Fetching is two MCP calls and sometimes three. The Actor tool returns when its own wait window elapses, not when the run finishes, so `status` can come back `RUNNING` over a dataset that exists and is empty. An agent that skips the poll reports products as not found, intermittently, depending only on how fast the retailer answered.
- Projecting with `fields` flattens the response into literal dotted keys, so reading the nested path finds nothing and looks identical to a missing field.
- Field names, types and nesting differ per retailer. Price is a number on one and a string on another; currency lives under a different key on a third; stock and rating are nested under `additionalProperties`. There is no unified schema and the skill says so.
- The `marketplaces` enum arrives truncated over MCP, so a supported retailer can be unselectable by keyword. Measured: of 249 values, 122 fit the server's cap and 127 were dropped.

**Limitations are stated, not omitted.** Measured latency over twenty questions is 14.7s fastest, 24.1s median, 95.8s slowest, so the skill says plainly that a runtime call is verification before the user acts rather than something to put in every turn.

**Validation.** `npm test` passes (14 tests, 0 failures) and `npm run validate` reports all skills valid. Branch follows `add-<component-name>`. One file added, nothing else touched: two files that changed only because I ran your tooling locally (`package-lock.json`, `hook-validation-report.json`) were reverted.

Tested by running the underlying Actor over MCP from Claude Code against real retailer pages, which is how each of the four gotchas above was found.

### DIFF
--- a/plugins/all-skills/skills/apify-product-lookup/SKILL.md
+++ b/plugins/all-skills/skills/apify-product-lookup/SKILL.md
@@ -1,0 +1,103 @@
+---
+name: apify-product-lookup
+description: "Fetch a real product's current price, stock, rating, or images from retailer pages over the Apify MCP server, and return them as typed fields rather than prose. Use when someone asks what a product costs right now, whether it is in stock, how it is rated, to compare a product across retailers, or to check a product URL they pasted. Also use when an agent would otherwise answer a product question from training data. Not for managing your own store's catalog, not for payments, and not for questions answerable without live retailer data."
+category: ecommerce
+license: MIT
+---
+
+# Apify product lookup
+
+A model answers product questions from training data that was fixed months ago, or by
+reading a page as prose with no product fields. This skill routes the question to a hosted
+Actor that reads retailer pages on request and returns structured records: price, currency,
+stock, rating, image URLs, and a link.
+
+Written by a product marketing manager at Apify. It calls
+[E-commerce Scraping Tool](https://apify.com/apify/e-commerce-scraping-tool), a paid
+first-party Apify Actor, so weigh the framing accordingly. Everything below is what
+testing against the live Actor returned, including where it does not work.
+
+## Setup
+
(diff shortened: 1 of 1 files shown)


davepoon:

**Security / prompt injection:** The skill markdown is operational, not override/exfil. Affiliation and the paid first-party Actor are disclosed in the intro (no mandatory promo footer). Setup points at the official Apify MCP (`https://mcp.apify.com`). The “never paste `additionalProperties` into a reply” rule is the right hardening for scraped marketing text. Frontmatter validates (`ecommerce`, name matches the directory).

**Codebase fit:** Distinct from the existing `msapps-apify-scraper` plugin (general Actor runner) and from the store-you-own ecommerce skills (`shopify-automation`, `stripe-automation`, `square-automation`). The two-step/three-step poll, flattened `fields`, and truncated `marketplaces` enum caveats are the useful part and belong in a catalog skill.

Merging.

PR #306, opened 2026-09-03
Update immich-photo-manager skill to 2.0.10

juanatsap:
Updates the immich-photo-manager skill page to the current version.

The page was written for 1.0 (21 tools, 11 skills) and the plugin is on 2.0.10 now: 94 MCP tools and 13 skills, OCR search, people and faces, metadata repair, video frames and PDF photobooks, notes on each asset between sessions, Immich 2.x and 3.x run live before every release, and a Docker image next to the plugin and uvx routes. The install block gets the `pip3 install -r src/requirements.txt` line the plugin needs since 1.7.1, and the credentials step after install.

Same file, same category (storage-docs), frontmatter unchanged except the description. `node scripts/validate-skills.js` passes on the branch.

### DIFF
--- a/plugins/all-skills/skills/immich-photo-manager/SKILL.md
+++ b/plugins/all-skills/skills/immich-photo-manager/SKILL.md
@@ -1,48 +1,69 @@
 ---
 name: immich-photo-manager
 category: storage-docs
-description: "Manage your self-hosted Immich photo library through conversation — natural language search, geographic album curation, duplicate detection, library health audits, and interactive HTML galleries. Install: claude plugin install immich-photo-manager"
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for the update — reviewed for security, prompt injection, and fit with this collection.

**Security / prompt injection:** Documentation-only change to an existing pointer skill. No override/exfil language. The install path still clones `drolosoft/immich-photo-manager` (MIT) and talks to a self-hosted Immich; nothing new leaves the user’s machine beyond what this skill already described. Schema validation and integration tests passed on the branch.

**Codebase fit:** Frontmatter `name`/`category` unchanged and valid. Feature counts and the `pip3 install -r src/requirements.txt` step match the current upstream plugin. You’re the original author of this listing, so this is the right place for the 2.0.10 refresh.

Merging.

PR #307, opened 2026-09-03
Add phaneslight plugin

Aloim:
## Summary

Adds **PhanesLight** to the marketplace: a multi-agent orchestration and project-memory system for Claude Code. One command turns a repository into a wired multi-agent environment with a project-specific agent roster, a documentation and session-summary tree, a tested script library, and enforcement hooks that hold size and stamp discipline at the harness layer. It is re-runnable rather than install-once, so the setup grows with the codebase instead of rotting beside it.

## Component details

| | |
|---|---|
| **Name** | `phaneslight` |
| **Type** | Plugin (two skills, three hooks, a shipped template library) |
| **Category** | `orchestration` |
| **Version** | 3.7.0 |
| **License** | See LICENSE (in-repo) |
| **Source** | `{"source": "github", "repo": "Aloim/phaneslight"}` |

Referenced by external repository rather than vendored under `plugins/`, matching the 13 existing `github`-source entries. The plugin ships a 220 KB skill body and a 40-file template library, so a vendored copy would drift from upstream on every release.

## Testing

- [x] **Validation passed.** `node scripts/validate-all.js` reports Hooks ✅ and Skills ✅. Subagents & Commands reports 512 errors, all in `plugins/all-commands/`, and they are **pre-existing on unmodified `main`**: identical count before and after this change, and none of them references `phaneslight`.
- [x] **Functionality tested.** `claude plugin validate ./plugin --strict` and `claude plugin validate . --strict` both pass on the source repository. A scratch-project end-to-end install completes at exit 0 with 27 files installed and the CLI smoke chain passing.
- [x] **No overlaps.** No existing entry named `phaneslight`; the entry is appended and no other line in the file is touched (28 insertions, 0 deletions).

## Usage examples

**1. Bootstrap a repository**

```
/phaneslight:run
```

Surveys the repository, infers module boundaries, generates a five-agent roster prefixed with the project slug, scaffolds `documentation/` and `tests/`, installs the script library, and writes the project memory. Pauses once to confirm module boundaries and once to ask whether to install the four recommended MCP servers.

**2. Keep an existing install current**

```
/phaneslight:run
```

The same command on an already-installed project runs as an update: it re-surveys, upgrades the sub-agents to the current lineup, fills in missing infrastructure, and bumps the run counter. A ledger lets a session that dies or compacts mid-run resume from the last completed phase.

**3. Migrate an older or manually installed project**

```
/phaneslight:upgrade
```

[...]

PR #309, opened 2026-09-04
Add phaneslight plugin (v3.7.2)

Aloim:
Follow-up to #307, which you closed pending another pass. Thanks for the review, all four points are addressed and the blocking one needed more than a repointed URL.

## Blocking: the source now resolves as a plugin

You were right that `Aloim/phaneslight` is the manual-install prompt pack. Repointing at the plugin repo alone would still not have worked, though, and that is worth saying plainly: a marketplace `github` source takes `repo`, `ref` and `sha` and **has no `path` field**, so Claude Code loads `.claude-plugin/plugin.json` from the referenced repository's **root**. My plugin sat under `plugin/`, reachable only from that repo's own `marketplace.json`, which is exactly why `/plugin marketplace add` worked while a listing here would have resolved nothing.

So two things changed upstream:

- The repository was renamed `Aloim/phanesplugin` to **`Aloim/phaneslightplugin`**, because `phanesplugin` reads as "the plugin for Phanes" while the product is PhanesLight, and that is what sent the previous PR at the manual repo in the first place. The old URL still redirects.
- **The repository is flattened.** `.claude-plugin/plugin.json` and the plugin content (`skills/`, `hooks/`, `scripts/`, `templates/`) are now at the root, and that repo's own marketplace entry points at `./` instead of `./plugin`. This matches the layout of the other `github`-source entries here.

`source.repo` in this PR is `Aloim/phaneslightplugin`, and `claude plugin validate` on this file passes; the 5 warnings it reports are all pre-existing and belong to other entries and the collection metadata.

## Version and license

- **`version` is `3.7.1`**, in sync with the published plugin. `plugin.json` and the entry agree.
- **`license` is the SPDX id `CC-BY-NC-4.0`.** The licence itself never changed; the field was prose.

## The two upstream notes

Both fixed in the plugin, not just noted:

**The fabricated-authority persona is gone.** The run skill opened by calling PhanesLight a laureate of an award and chief architect of an institute, neither of which exists, and asked for "world-class, award-winning" sub-agents. That is cut, replaced by the plainer wording my v4.0 line already uses. No directive, phase or threshold moved, it is register only.

**The hooks are no longer PowerShell-only.** This turned out to be a regression rather than a Windows-first limitation, and the diagnosis is the useful part: before the plugin packaging, the setup run merged hook entries into each project's own `.claude/settings.json`, which happens on the target machine, so it could write a `powershell` entry on Windows, a `sh` entry on POSIX, and omit an entry entirely where no variant existed. Moving registration into the plugin's single static `hooks.json` took that choice away without replacing it, and the shipped file was PowerShell everywhere.

Entries now run `node "${CLAUDE_PLUGIN_ROOT}/scripts/hook-run.js" <hook-name>`. The launcher resolves the platform's own script extension, dispatches through `powershell` on Windows and `sh` elsewhere, and passes the hook's exit code through unchanged so the blocking `PreToolUse` guard still blocks. A hook with no variant on this platform is a silent no-op rather than a broken command, which is also what all of them are in a project that has no `.phaneslight/` at all. `install-notice.ps1` is ported to Node outright. Verified against a scratch tree: a child exit of 7 arrived as 7, a missing variant exited 0 silently, and the hook JSON payload reached the child's stdin intact.

One thing I want to be upfront about rather than have you find it: **part of the script library PhanesLight installs into a project is still Windows-only** (`preflight`, `scaffold`, `install-templates`, `ledger`, `manifest-write`, `census-diff`, `update-preflight`, `repo-manifest`, `batch-apply`, and the `hook-ledger-status` hook). On macOS and Linux the run does that work itself instead, and the day-to-day checks (`doc-check`, `doc-index`, `loc-check`, `register-check`, `new-file`, `module-list`, `list-apis`, and both enforcement hooks) are on both platforms. The README states this by name now instead of saying "Windows-first" and leaving a reader to guess what they lose.

Happy to take another pass on anything.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -2170,6 +2170,34 @@
         "repo": "JasonColapietro/suede-creator-skills"
       },
       "strict": false
+    },
+    {
+      "name": "phaneslight",
+      "description": "Multi-agent orchestration and project-memory system for Claude Code. One command turns a repository into a wired multi-agent environment: a project-specific agent roster, a documentation and session-summary tree, a tested script library, and enforcement hooks that hold size and stamp discipline at the harness layer. Re-runnable, so the setup grows with the codebase.",
+      "author": {
+        "name": "Aloim",
+        "url": "https://github.com/Aloim"
+      },
+      "homepage": "https://github.com/Aloim/phaneslightplugin",
+      "repository": "https://github.com/Aloim/phaneslightplugin",
+      "license": "CC-BY-NC-4.0",
+      "keywords": [
+        "orchestration",
+        "multi-agent",
+        "subagents",
+        "project-memory",
+        "documentation",
+        "hooks",
+        "bootstrap",
+        "code-quality",
+        "claude-code"
+      ],
+      "category": "orchestration",
+      "version": "3.7.1",
+      "source": {
+        "source": "github",
+        "repo": "Aloim/phaneslightplugin"
+      }
     }
   ]
 }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -2193,7 +2193,7 @@
         "claude-code"
       ],
       "category": "orchestration",
-      "version": "3.7.1",
+      "version": "3.7.2",
       "source": {
         "source": "github",
         "repo": "Aloim/phaneslightplugin"


Aloim:
Updated the entry from 3.7.1 to 3.7.2.

The source is unpinned (no `ref`, no `sha`), so it resolves the default branch, and v3.7.2 is now what a collection install actually delivers. Leaving the entry at 3.7.1 would have misreported the version rather than served an old one.

**What changed in v3.7.2:** full platform parity. Every script the plugin installs into a project now has both a Windows and a POSIX variant. The v3.7.1 submission noted that ten of them were Windows-only with the run falling back to doing that work itself; that gap is closed. Seven of the ten use Node to parse JSON, which adds no dependency, since Claude Code is itself a Node program and the plugin already dispatches its hooks through Node.

No other field changed, and nothing about the earlier review findings has regressed: `license` is still the SPDX id `CC-BY-NC-4.0`, the hooks still dispatch cross-platform through a launcher, and `source.repo` still points at the flattened plugin repository.

Release notes: https://github.com/Aloim/phaneslightplugin/releases/tag/v3.7.2

davepoon:
Thanks for the follow-up — this addresses the #307 blockers.

**Codebase fit:** `source.repo` now points at `Aloim/phaneslightplugin`, which has `.claude-plugin/plugin.json` at the repo root (`version` `3.7.2`, matching this entry). That is the layout a github-source listing here actually loads. `license` is the SPDX id `CC-BY-NC-4.0`. Category `orchestration` matches other github-source plugins in this file.

**Security / prompt injection:** Hook dispatch is a Node launcher with an allowlisted script name and `spawnSync` (list argv, no shell). POSIX and Windows variants are both present. The run skill is user-invoked (`disable-model-invocation: true`); the fabricated award/institute persona from the earlier sample is gone. Remaining “world-class expert” language is spawn-persona framing, not fake credentials for the product. Secret-handling guidance in the run skill is solid.

Merging.

PR #286, opened 2026-08-21
Add video-to-text, de-ai-writer, ecommerce-material-studio (schema-valid frontmatter fix)

jiawood2006:
Follow-up PR: de-ai-writer frontmatter now uses full-width quotes (yaml.safe_load validates cleanly). Replaces closed PR #282. Skills: video-to-text, de-ai-writer, ecommerce-material-studio — collection-relative scripts, no donation footer, valid categories.

PR #310, opened 2026-09-05
Add voice-persona: two-way WeChat voice (silk STT + persona TTS)

jiawood2006:
Adds **voice-persona** — the first IM two-way voice skill: transcribe voice messages (incl. WeChat .silk) and reply in persona voices encoded back to WeChat voice bubbles.

**What it does**
- `stt` — any audio file → text; WeChat `.silk` supported (pilk; most open solutions can't decode silk)
- `speak` — text → persona-flavored reply → Edge TTS Chinese voice (6 personas built in)
- `to_silk` — audio → WeChat voice bubble format (`#!SILK_V3`)
- `demo` — 30s no-key smoke test of the full loop
- All local / free: faster-whisper + pilk + edge-tts, zero API cost

**Diff vs existing voice skills**: others do one-way "voice→text" (jozhn/wechat-voice-decode-skill) or cloud ASR/TTS with ban-risk (chatgpt-on-wechat); voice-persona closes the two-way voice loop in IM with persona voices.

Files: `plugins/all-skills/skills/voice-persona/` (SKILL.md + scripts/voice_persona.py) — same layout as previously merged PR #286.

### DIFF
--- a/plugins/all-skills/skills/voice-persona/SKILL.md
+++ b/plugins/all-skills/skills/voice-persona/SKILL.md
@@ -1,14 +1,8 @@
 ---
 name: voice-persona
 description: "让 Agent 变成能语音对话的机器人：语音文件转文字（支持微信 silk 格式）+ 多音色人格回复（Edge TTS 免费中文音色），全本地零 API 成本。Voice persona chat: transcribe voice messages (incl. WeChat silk) and reply in persona voices."
-version: 1.0.0
-author: 涛哥
+category: communication
 license: MIT
-metadata:
-  hermes:
-    tags: [voice, stt, tts, persona, silk, wechat, whisper, edge-tts, speech, 语音, 人格]
-    category: autonomous-ai-agents
-    homepage: https://github.com/jiawood2006/hermes-skills
 ---
 
 # Voice-Persona · 语音对话 + 人格音色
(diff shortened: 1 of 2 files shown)


davepoon:
Thanks for shipping a real script alongside the skill — silk STT + persona TTS is a useful loop, and the Python stays off `shell=True`.

**Security / prompt injection:** No classic malice. Subprocess calls are list-argv. Optional `--llm` follows the same `LLM_API_KEY` / OpenAI-compatible URL pattern as skills already in this collection. Edge TTS being a network call is documented. The markdown does not try to override the harness.

**Codebase fit — blocking:** the documented commands do not work from this collection.

Skills here invoke helpers relative to the skill folder (see `video-to-text` / `de-ai-writer`: `python3 scripts/…`). This PR vendors `scripts/voice_persona.py` correctly, but SKILL.md tells the agent to:

- run `python3 skills/voice-persona/scripts/voice_persona.py demo` (Hermes repo layout, not this tree)
- `hermes skills install …` / `git clone https://github.com/jiawood2006/hermes-skills`

An agent following those steps will miss the files that actually shipped in the PR. Schema/CI are fine; the install and smoke-test paths are what fail the bar.

- uses skill-relative commands (`python3 scripts/voice_persona.py …`) as the primary path
- keeps Hermes as an optional note, not the install instructions
- leaves the vendored script in place

Appreciate the contribution — closing for now so it’s clear this needs another pass. Please open a fresh PR (or reopen after fixing) once the usage path matches this collection.

PR #311, opened 2026-09-06
Add voice-persona: two-way WeChat voice (silk STT + persona TTS)

jiawood2006:
Adds **voice-persona** — a voice chat skill with a real, vendored script.

**What it does:** voice → text (incl. WeChat .silk via pilk) → persona reply (6 built-in Chinese personas) → Edge TTS voice → optional .silk encode for IM voice bubbles. Fully local / zero API cost (faster-whisper + Edge TTS + pilk), optional `--llm` for tone polish.

**Follow-up to #310 (closed):** usage paths now follow this collection's convention — commands are skill-relative (`python3 scripts/voice_persona.py …`, matching `video-to-text` / `de-ai-writer`). Hermes install is an optional note, not the primary path. Vendored script unchanged.

**Security:** subprocess calls are list-argv (no `shell=True`); optional `--llm` follows the same `LLM_API_KEY` / OpenAI-compatible URL pattern as existing skills; Edge TTS network call documented; SKILL.md does not override the harness.

Smoke test verified from the skill folder:
```bash
cd plugins/all-skills/skills/voice-persona
python3 scripts/voice_persona.py demo
```
→ STT → 3 persona TTS mp3s → .silk encode all pass.

### DIFF
--- a/plugins/all-skills/skills/voice-persona/SKILL.md
+++ b/plugins/all-skills/skills/voice-persona/SKILL.md
@@ -1,14 +1,8 @@
 ---
 name: voice-persona
 description: "让 Agent 变成能语音对话的机器人：语音文件转文字（支持微信 silk 格式）+ 多音色人格回复（Edge TTS 免费中文音色），全本地零 API 成本。Voice persona chat: transcribe voice messages (incl. WeChat silk) and reply in persona voices."
-version: 1.0.0
-author: 涛哥
+category: communication
 license: MIT
-metadata:
-  hermes:
-    tags: [voice, stt, tts, persona, silk, wechat, whisper, edge-tts, speech, 语音, 人格]
-    category: autonomous-ai-agents
-    homepage: https://github.com/jiawood2006/hermes-skills
 ---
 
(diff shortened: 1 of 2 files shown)


davepoon:
Thanks for the follow-up — the usage path now matches this collection.

Reviewed for security, prompt injection, and fit.

**Security / prompt injection:** Still clean. Subprocess calls stay list-argv (no `shell=True`). Optional `--llm` is opt-in via `LLM_API_KEY` and does not override the harness. Edge TTS as a network call is documented. SKILL.md has no instruction-override / exfil language.

**Codebase fit:** Schema/CI pass. Primary commands are skill-relative (`python3 scripts/voice_persona.py …`), the script is vendored, and Hermes is an optional note. That was the gap on #310.

Merging.

jiawood2006:
Thanks for the careful review and the merge! The feedback on #310 (skill-relative commands, vendored scripts, keeping Hermes as an optional note) was spot-on — glad the revised structure fits the collection now. Happy to iterate further if any integration issues come up on your side.

PR #312, opened 2026-09-08
Add you-web-search skill

mouse-value-add:
## Summary

Adds a `you-web-search` skill that gives Claude Code current web search and URL reading through the You.com MCP server. It follows the catalog's MCP-backed skill pattern (like `coinpaprika-api`): one SKILL.md that documents the remote MCP endpoint, setup, and when the model should reach for it.

## Component Details
- **Name**: you-web-search
- **Type**: Skill
- **Category**: research

The server is remote, so setup is a single `mcpServers` entry — no local install. Two options are documented:
- keyless basic search: `https://api.you.com/mcp?profile=free`
- full tools with an API key: `https://api.you.com/mcp` + `Authorization: Bearer ${YDC_API_KEY}` (key from https://you.com/platform/api-keys)

There's currently no web-search skill in the catalog backed by a remote MCP server, so this fills a gap for questions where training data is stale.

## Testing
- [x] Ran validation (`npm test`) — skills schema, subagents, hooks, and all 14 unit tests pass
- [x] Tested functionality — MCP endpoints verified live (keyless `you-search` responds over the `profile=free` endpoint)
- [x] No overlap with existing components — searched the catalog; no You.com or remote-search MCP skill exists

## Examples

- "What's the current stable version of Node.js?" → `you-search`, answer with source URLs
- "Find recent posts about the React compiler and summarize the reception" → search, extract, cite
- "Check the latest docs for this API before writing the integration" → search + URL reading

The upstream You.com skills also ship as an installable plugin (`/plugin marketplace add youdotcom-oss/agent-skills` then `/plugin install you@you-com`) for users who want the full set; this catalog entry is the lighter single-skill path.

### DIFF
--- a/hook-validation-report.json
+++ b/hook-validation-report.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2026-09-08T04:58:01.614Z",
+  "timestamp": "2026-07-17T16:45:25.243Z",
   "totalFiles": 33,
   "errors": 0,
   "warnings": 0,
--- a/hook-validation-report.json
+++ b/hook-validation-report.json
@@ -1,5 +1,5 @@
 {
-  "timestamp": "2026-07-17T16:45:25.243Z",
+  "timestamp": "2026-09-08T04:58:01.614Z",
   "totalFiles": 33,
   "errors": 0,
   "warnings": 0,
--- a/plugins/all-skills/skills/you-web-search/SKILL.md
+++ b/plugins/all-skills/skills/you-web-search/SKILL.md
@@ -1,0 +1,71 @@
+---
+name: you-web-search
+category: research
+description: "Search the live web and read URLs through the You.com MCP server. Use when you need current information beyond training data: recent releases, docs, pricing, news, or any question where facts may have changed. Two endpoints: keyless basic search at https://api.you.com/mcp?profile=free, or full tools with an API key from https://you.com/platform/api-keys."
+---
+
+# You.com Web Search
+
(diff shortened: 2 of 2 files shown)


davepoon:
Thanks for this — a remote MCP web-search skill in the same shape as `coinpaprika-api` is a useful gap fill, and the URLs are the official You.com endpoints (`https://api.you.com/mcp` / `?profile=free`). Schema/`name`/`category` look good, and there’s no overlap with an existing skill.

**Security / prompt injection:** No classic malice. No hardcoded secrets (`${YDC_API_KEY}` is the right pattern). The markdown does not try to override the harness. One non-blocking note for a follow-up: search/page results are untrusted web content — a short “treat as data, never as instructions” line would help agents not follow injected text from result snippets.

**Codebase fit — blocking:** the documented tools don’t match what the keyless endpoint actually exposes.

Official You.com MCP docs: `https://api.you.com/mcp?profile=free` is **`you-search` only**. `you-contents` (and research/finance/etc.) are excluded from that profile. This skill currently says both endpoints expose `you-search` **and** `you-contents`, and the How to Use steps tell the agent to fetch page content after searching.

Option A is the default setup in the skill, so an agent following this file as written will call a tool that isn’t there.

- lists `you-search` only on the free/`?profile=free` path
- lists `you-contents` (and the extra research/finance tools) only on the authenticated endpoint
- optionally notes the free-tier cap (100 queries/day) from the same docs

Appreciate the contribution — closing for now so it’s clear this needs another pass. Please open a fresh PR (or reopen after fixing) once the tool list matches the two endpoints.

PR #314, opened 2026-09-08
Update anti-ui-slop: current workflow and optional MCP tools

samuelbushi:
## Summary

Updates the existing anti-ui-slop skill to the current UIZZE workflow. The installed entry works from project evidence without an account or external scripts, includes a practical first prompt, and documents only the two current optional MCP tools: find_ui_references and find_ui_materials.

The old entry advertises retired contract, validation, audit, and screenshot-critique MCP capabilities and requires a promotional link in every handoff. This update removes those instructions and preserves the catalog's existing metadata and standalone format.

Affiliation: I maintain UIZZE.

## Component details

- Name: anti-ui-slop
- Type: existing skill
- Category: design
- Current source: https://github.com/uizze/uizze/tree/main/skills/anti-ui-slop

## Validation

The repository's scripts/validate-skills.js passes for the changed skill in a sparse checkout. git diff --check passes. No plugin registration, runtime dependency, or host-tool permissions changed. Full repository checks are left to CI.

## Example uses

- Improve billing settings with invoice, payment-failure, and read-only states.
- Review a generic first draft and implement the three most useful product-specific changes.
- Use connected MCP references for a concrete permissions or navigation question.

### DIFF
--- a/plugins/all-skills/skills/anti-ui-slop/SKILL.md
+++ b/plugins/all-skills/skills/anti-ui-slop/SKILL.md
@@ -4,67 +4,52 @@ category: design
 description: Stop coding agents from shipping generic UI. Use UIZZE's 800,000+ real web and iOS screens to build product-specific interfaces, define a design contract, cover required states, and run a hard finish gate. Use for web or iOS UI design, implementation, redesign, critique, and pre-ship review in Codex, Claude Code, Cursor, Copilot, and other coding agents.
 ---
 
-> ***If your UI screams AI, your app is dead.***
-
 # Stop Making UI Slop
 
-Build distinctive UI with 800,000+ real web and iOS screens via [UIZZE](https://uizze.com).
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for the update — this is a cleaner catalog entry than what we have now, and it matches the current UIZZE workflow (standalone, no account required; paid MCP is optional and limited to `find_ui_references` / `find_ui_materials`).

**Security / prompt injection:** No classic malice. The skill does not try to override the harness, does not harvest credentials, and no longer requires a promotional link in every handoff. Optional MCP use is gated on a real host connection, and empty retrievals are handled honestly (“continue from the project,” never invent a tool result). Schema/`name`/`category` are unchanged and valid.

**Codebase fit:** The file stays in the catalog’s standalone `SKILL.md` shape (no extra frontmatter that would fail validation). The workflow is usable from repository evidence alone, which is the right default here.

**Note (non-blocking):** The existing description still mentions a “hard finish gate” and treating the 800k-screen catalogue as the core path, while the body now makes those optional/paid. Happy to take a small follow-up that trims the description to match.

Merging.

PR #322, opened 2026-09-12
Add orca-replay skill

xizhuomengcontin:
Adds one skill at `plugins/all-skills/skills/orca-replay/SKILL.md`, in the structure
CONTRIBUTING specifies (`name` / `category` / `description` front matter, then When to Use → What
This Skill Does → How to Use → Example → Tips).

### What it is

A skill that stops the agent answering questions about its own past from memory. When someone asks
"why did you overwrite that file" or "which step broke the build", it reads the recording of the
run — tool calls with their arguments, shell exit codes, the files that changed — and answers from
that instead of reconstructing a plausible story.

The part I care most about: the trace labels every causal edge `recorded` or `inferred`, and the
skill requires the agent to carry that distinction into its answer. *"The trace shows the `rm` at
step 14 removed it"* and *"this looks like the `rm` at step 14, going by timing"* are different
claims, and flattening them into one confident sentence is the failure the skill exists to prevent.

### Two things you should know before merging

**It depends on an external tool.** The skill needs the `orcareplay` npm package (Node 20+) with
its MCP server registered as `orca`, plus at least one recorded run in the project. I have put that
in a **Requirements** section at the end rather than burying it, and the skill's own instruction
when no recording exists is to say so and offer to start one — not to fall back to guessing. If
your marketplace would rather not carry skills that require an outside install, that is a fair
line and I will withdraw this.

**It is a copy, and copies go stale.** The canonical skill lives at
[`skills/orca-replay/SKILL.md`](https://github.com/Continuum-AI-Corp/OrcaReplay/blob/main/skills/orca-replay/SKILL.md)
in our repo. This version is condensed to your format; the upstream one is longer, mostly because
it spells out replay's destructive cases at more length. I am happy to keep this in sync, and
happier still if you would rather point at the source than vendor it.

Checked: `orca-replay` does not overlap an existing skill in `plugins/all-skills/skills/`, one
purpose per contribution, Apache-2.0.

### DIFF
--- a/plugins/all-skills/skills/orca-replay/SKILL.md
+++ b/plugins/all-skills/skills/orca-replay/SKILL.md
@@ -1,0 +1,93 @@
+---
+name: orca-replay
+category: development
+description: Answers questions about a past agent run from its recording rather than from memory, and replays or forks that run. Use when asked why an earlier run did something, or to reproduce a failure.
+---
+
+# OrcaReplay Run Reader
+
+Coding agents are confidently wrong about their own history. A summary written from memory is
+missing the tool results, the exit codes, and the files that changed without anyone mentioning
+them.
+
+This skill enforces one rule: **when a question is about something that already happened, read the
+trace before answering.** Do not reconstruct it. If a recording exists, guessing is the wrong move
+even when the guess would have been right.
+
+## When to Use This Skill
+
+- "Why did you delete / overwrite / move X?"
+- "What changed this file?" or "Which step broke the build?"
+- "Can you reproduce yesterday's failure?"
+- "Would a different model have got this right?"
+
+## What This Skill Does
+
+1. **Finds the run** — `orca_list_runs`, newest first, naming the run each fork came from.
+2. **Narrows to the cause** — `orca_graph` with `to: <event seq>` returns only the chain that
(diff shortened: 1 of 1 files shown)


davepoon:
Thanks for sending this over — the recorded-vs-inferred distinction is a good rule, and the replay warnings (worktree, no sandbox, matching replay ≠ determinism) are honest.

**Security / prompt injection:** No classic malice in the skill text. Requiring the agent to answer from the trace instead of memory is instruction, not injection. The filesystem-overwrite risk on `orca_replay` without `worktree: true` is documented clearly.

**Codebase fit — blocking:**

1. **Invalid skill category** — `category: development` is not in the skill enum. Skills in `plugins/all-skills/` must use one of the values in `scripts/skill-schema.json` (for this one, `development-code` is the match). Validation will fail as written, same bar as other category mismatches.

The external `orcareplay` MCP requirement is fine to keep if it stays in a **Requirements** section (other skills here depend on an outside install). Happy to re-review a follow-up with `category: development-code`.

Appreciate the contribution — closing for now so it’s clear this needs another pass. Please open a fresh PR (or reopen after fixing) once the category is valid.

PR #326, opened 2026-09-15
Add orca-replay skill (category: development-code)

xizhuomengcontin:
Follow-up to #322, as invited: *"Happy to re-review a follow-up with `category: development-code`."*

**Disclosure: I maintain OrcaReplay.** Apache-2.0, free, no paid tier, no account.

## The one change

```diff
-category: development
+category: development-code
```

`development` is not in the enum. I checked `scripts/skill-schema.json` rather than taking the value on trust — `/properties/category` lists 27 values, and `development-code` is the one you named. The file is otherwise byte-identical to what you reviewed in #322.

## Everything else, left as reviewed

- **Requirements section kept where it was** (`## Requirements`, line 89), since you said an external install is fine as long as it stays there.
- The recorded-vs-inferred rule, the worktree warning, the "replay is not a sandbox" note and the "a matching replay is not a determinism result" line are all unchanged — you called those out as the parts worth keeping, so I did not touch them while I was in the file.

## Why a fresh PR rather than a reopen

You offered either. A fresh one seemed cleaner than reopening a branch whose head now differs from what you reviewed; this way the diff you see is the file as it will land. Say the word if you would rather have had #322 reopened and I will close this.

Nothing else to flag — the blocking item was one line and this is it.


davepoon:
Thanks for the follow-up — category is now a valid skill enum value (`development-code`), which was the only blocking item from #322.

**Security / prompt injection:** No classic malice in the skill text. Requiring the agent to answer from the trace (and to keep `recorded` vs `inferred` distinct) is instruction, not injection. The filesystem-overwrite risk on replay without `worktree: true` is still documented clearly. External `orcareplay` MCP stays in **Requirements**, which matches how other skills here depend on an outside install.

Schema/`name`/directory match look good, and there’s no overlap with an existing skill.

Merging.

PR #331, opened 2026-09-20
Add isitdone plugin

raimondasl:
## Summary

Adds [isitdone](https://github.com/raimondasl/isitdone) to the plugin marketplace. It is a Stop hook for Claude Code that runs the repository's real test, typecheck and lint commands on the exact working tree when Claude tries to end a turn claiming the work is done, and blocks the turn with the actual failure output until they pass (three attempts, then it lets Claude stop and says why). A PostToolUse hook warns right after an edit that weakened a test, and the bundled skill teaches Claude to run `npx isitdone` before claiming completion. No LLM calls, no network, zero dependencies, MIT.

One entry appended to `.claude-plugin/marketplace.json`, using the external-source form that curviate and other entries already use, so nothing is vendored here and the listing tracks the upstream repository. The fields are copied from the plugin's own `.claude-plugin/plugin.json`.

## Component Details
- **Name**: isitdone
- **Type**: Plugin (Stop hook, PostToolUse hook, one skill)
- **Category**: development

## Testing
- [x] Ran validation (`npm run validate`): the marketplace file parses and the entry count goes from 89 to 90. The run reports 512 errors under `plugins/all-commands` and `plugins/all-agents`; the same 512 appear on an untouched checkout of `main`, so they are not from this change.
- [x] Tested functionality: `/plugin marketplace add raimondasl/isitdone` then `/plugin install isitdone@isitdone`; `claude plugin validate . --strict` passes in the plugin repository.
- [x] No overlap with existing components: the existing hooks here notify, format or lint; none gates the end of a turn on the repository's own checks.

## Examples
- Claude ends a refactor with "Done, all 48 tests pass." The hook runs `npm test`, one test fails, and Claude receives the failing assertion and fixes it before the turn can end.
- Claude adds `it.skip` to get a suite green. The PostToolUse warning names the line; with `"integrity": "strict"` the stop is blocked until the test is restored.
- `npx isitdone history` grades past "done" claims from local Claude Code transcripts (verified, stale, failed, never ran); nothing leaves the machine.

Disclosure: isitdone is built and maintained by Claude under my direction, as an experiment in agent-driven open source; I approve releases.


davepoon:

**Security / prompt injection:** The marketplace entry itself is metadata-only. I also checked upstream `raimondasl/isitdone` and the published `@aivolution/isitdone@0.5.1` package: zero runtime dependencies, no outbound HTTP client in the bundle, and the Stop / PostToolUse hooks stay on the local working tree (spawn the repo’s own test/typecheck/lint commands, write `.isitdone/` receipts). The skill text does not try to override host instructions. Complementary to `red-handed-audit` (transcript audit vs actually running the repo checks), so not a duplicate.

**Codebase fit:** Same external-source form as `curviate` / `archcore` / `forge`, fields match the plugin’s own manifest, and `category: development` is already used here.

**Notes (non-blocking):**
- Catalog `version` is `0.5.0`; upstream `plugin.json` is now `0.5.1`. Happy to take a follow-up bump.
- The hook command uses unpinned `npx -y @aivolution/isitdone`. Pinning a version would reduce supply-chain drift, same note we have made on similar listings.
- The skill currently lives at repo-root `SKILL.md` (skills.sh / `npx skills add`). Claude Code plugin discovery looks under `skills/`, so `/plugin install` ships the hooks; the skill is a separate install path.

Merging.

PR #334, opened 2026-09-20
Bump isitdone to 0.5.2

raimondasl:
Follow-up to #331, as offered in the review: the catalog entry goes from 0.5.0 to 0.5.2, matching upstream `plugin.json`.

0.5.2 also acts on the other two review notes, thank you for them:
- The plugin's `hooks/hooks.json` now pins the npm version it runs (`npx -y @aivolution/isitdone@0.5.2 hook ...`), so a plugin at version X always runs package X.
- The skill now also ships under `skills/isitdone/SKILL.md`, so `/plugin install` delivers it; the root `SKILL.md` stays for `npx skills add`.


davepoon:
Thanks for the follow-up — reviewed for security, prompt injection, and fit with this collection.

**Security / prompt injection:** Metadata-only catalog bump. Confirmed on the `v0.5.2` tag: `hooks/hooks.json` pins `npx -y @aivolution/isitdone@0.5.2`, and the skill now also ships at `skills/isitdone/SKILL.md`. The published package still has zero runtime deps and no outbound HTTP client; the skill text does not try to override host instructions.

**Codebase fit:** This is the version/pin/skill-path follow-up offered on #331. CI green.

**Note (non-blocking):** Upstream `plugin.json` / hooks are already on `0.6.0`. The GitHub source still installs from the repo; a later bump would keep the listing in sync.

Merging.

PR #335, opened 2026-09-21
Bump archcore to 0.9.3 (sync with upstream plugin.json)

ivklgn:
Hi @davepoon! This bumps the Archcore entry to **v0.9.3** and syncs it with the upstream `plugin.json` (https://github.com/archcore-ai/plugin/releases/tag/v0.9.3).

- `.claude-plugin/marketplace.json` → `archcore`:
  - `version` 0.4.4 → 0.9.3, `description` synced
  - `source` now points at the plugin folder: `git-subdir` with `path: plugins/archcore`. The repo root of `archcore-ai/plugin` holds our own marketplace, not the plugin, so the old `github` source installs a plugin with no skills, hooks or agents. I checked both in a clean `CLAUDE_CONFIG_DIR`: the repo root installs 0 skills, the `git-subdir` entry installs all 4 (`init`, `plan`, `document`, `review`). The official Anthropic marketplace uses the same `git-subdir` form for plugins in a subfolder.
- `plugins/archcore/.claude-plugin/plugin.json`: `version` and `description`, same values.
- `plugins/archcore/README.md`: rewritten for the current plugin. The old one listed seven commands that no longer exist (`context`, `capture`, `decide`, `audit`, `help`); the plugin now has four (`init`, `plan`, `document`, `review`) and also runs in GitHub Copilot.

I maintain Archcore and send these bumps only for minor and major releases. If you'd rather not get them, tell me and I'll stop.


davepoon:
Thanks for the bump. Reviewed for security, prompt injection, and fit with this collection.

**Security:** Checked the v0.9.3 plugin under `plugins/archcore`. Hooks are thin launchers: they refuse to run from a plugin-cache directory, fail open when the CLI is missing or too old, and hand the payload to `archcore` on `PATH`. The hook scripts do not fetch the network or eval input. MCP is the local `archcore mcp` process. Init grounding explicitly says to record secret names and purpose only, never values. The `curl | bash` install line is the same CLI install path this listing already documented.

**Prompt injection:** Skills, agents, and commands are project-context instructions. No override, credential-harvest, or exfil patterns in the shipped text.

**Codebase fit:** Pointing `source` at `plugins/archcore` is the right fix. The repo root is a marketplace, so the old `github` source installed a plugin with no skills. Version, description, and README now match the four commands that actually ship (`init`, `plan`, `document`, `review`). The local stub under `plugins/archcore/` stays in sync with the marketplace entry.

One optional follow-up: `ref` is `main`, so installs track the branch rather than tag `v0.9.3`. The previous entry was also unpinned. Pinning the tag would make the version label match the bytes users get.

Merging.

